PR #18, opened 2017-05-23
Test Forte w/o building Psi4

loriab:
I wouldn't accept this now unless you're really keen configuring Travis tonight, and in any case this should be squashed. It is working to the extent that Forte is running against pre-built Psi4 and freshly compiled ambit. All the tests that look to be running are passing, but the CASSCF and a few others are failing. Any chance these are expected failures?

https://travis-ci.org/loriab/forte/builds/235067991?utm_source=github_status&utm_medium=notification

Unfortunately using Psi4 pre-built with gcc-5.2 constrains Ambit and Forte, so the Travis build matrix isn't so useful.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,27 +26,51 @@ matrix:
 #      - BUILD_TYPE='release'
 #      - NAME='gcc'
 #      - VERSION='4.9'
+#  - os: linux
+#    compiler: clang
+#    addons: &1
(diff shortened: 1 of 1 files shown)


fevangelista:
Thanks for this PR. I am not sure why the ECP and CASSCF tests are failing. Right now have not clue how to go about it. Let me think...

PR #19, opened 2017-05-24
Travis working w/o psi4 build. Intro pure CMake build.

loriab:


fevangelista:
Thanks @loriab! This is the first external PR for Forte!

PR #23, opened 2017-05-26
CheMPS2, GlobalArrays, pytest

loriab:
- [x] Includes a sketchy GlobalArrays locator machinery via CMake. Untested. GlobalArrays does, in the newest version, build alternately through CMake, but the min CMake they require is too old to allow imported targets. I think GA is sufficiently remote from us to not mess with their build system and contribute back.
- [x] Allows CheMPS2 to be optional. Do `cmake ... -DENABLE_CheMPS2=ON` to require it. I did version `1.8.3` (same as Psi4) more for the build version than the contents version.
- [x] Introduces pytest. After build, just `pytest -v` and you should get 4 passes or 2 passes/2 skipped on psi4 1.1 (no ecp)
   * Requires rewriting tests in PsiAPI mode (straightforward but tedious). Files can contain multiple tests. Most appropriate use of file grouping is to quickly test subset (`pytest -v test_aci.py`)
   * To work, requires Psi4 to have [these two lines](https://github.com/psi4/psi4/blob/master/psi4/src/export_plugins.cc#L66-L67) _outside_ the `if` statement. Otherwise only first-run forte test will work.
- [x] Using `PLUGIN_SOFILE` from `__init__py` as a global module variable

@jturney

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,7 @@
 *.pyc
 Makefile.*
 Makefile
+timer.dat
 *~
 
 # Qt Creator project files
@@ -25,6 +26,7 @@ Makefile
 /tests/methods/*/timer.dat
 /tests/methods/*/*.txt
 /tests/methods/*/*.npz
+tests/pytest/.cache/
 
 doc/doxygen/html
 doc/doxygen/latex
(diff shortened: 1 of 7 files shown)


fevangelista:
Thanks for this contribution @loriab!

fevangelista:
Just ran the pytests and work exactly as you described it.

loriab:
Good to hear. `pytest -s` is also a handy argument – it allows stdout from the tests through to the screen.

PR #41, opened 2017-08-05
Integrate density fitting into MRDSRG.

tyzhang1993:
This pull request integrates density fitting into MRDSRG.

# Goals #
## Add DF capabilities ##
- [x] Modify code to generate V from density fitted basis.
- [x] Added density fitted test cases.

## Implement sequential MRDSRG algorithm ##
- [x] Implement orbital rotation approach

## Commutators approximation ##
- [x] Omit V3 blocks

# Status #
- [x] Ready to go.


tyzhang1993:
@lcyyork Thank you so much! Working on it now.

src/mrdsrg-spin-integrated/mrdsrg_nonpt.cc
```diff
@@ -384,9 +384,23 @@ void MRDSRG::compute_hbar_sequential_rotation() {
     H1 = BTF_->build(tensor_type_, "Transformed H1", spin_cases({"gg"}));
     H1["rs"] = U1["rp"] * H_["pq"] * U1["sq"];
     H1["RS"] = U1["RP"] * H_["PQ"] * U1["SQ"];
-    Hbar2_["pqrs"] = U1["pt"] * U1["qo"] * V_["to12"] * U1["r1"] * U1["s2"];
-    Hbar2_["pQrS"] = U1["pt"] * U1["QO"] * V_["tO19"] * U1["r1"] * U1["S9"];
-    Hbar2_["PQRS"] = U1["PT"] * U1["QO"] * V_["TO89"] * U1["R8"] * U1["S9"];
+    if (eri_df_) {
+        ambit::BlockedTensor B;
+        B = BTF_->build(tensor_type_, "B 3-idx", {"Lgg", "LGG"});
```
fevangelista: In future we might want to avoid allocating/deallocating this tensor.

src/mrdsrg-spin-integrated/mrdsrg_nonpt.cc
```diff
@@ -512,23 +527,26 @@ void MRDSRG::compute_hbar_sequential_rotation() {
...
-        TIME_LINE(Hbar2_["pQrS"] = O2_["pQrS"]);
-        TIME_LINE(Hbar2_["PQRS"] = O2_["PQRS"]);
+        TIME_LINE(Hbar2_["pqrs"] = B["gpr"] * B["gqs"]);
+        TIME_LINE(Hbar2_["pqrs"] -= B["gps"] * B["gqr"]);
+
+        TIME_LINE(Hbar2_["pQrS"] = B["gpr"] * B["gQS"]);
+
+        TIME_LINE(Hbar2_["PQRS"] = B["gPR"] * B["gQS"]);
+        TIME_LINE(Hbar2_["PQRS"] -= B["gPS"] * B["gQR"]);
+
```
lcyyork: Hey Sam, great work! One thing for optimization: if we want to store O2 with at most two virtual indices, it might be better to store all those blocks from the very beginning and only density fitted the bare H with three and four virtual indices. As such we do not contract B for each term, although you might want to time which one is faster.

tyzhang1993: Thank you for the suggestion! That's a good idea. I am actually working on V_{"gggg"}, which actually can be one way to do it.

PR #50, opened 2018-01-04
Move DWMS-DSRG-PT2 code to a class.

lcyyork:
- [x] Move DWMS-DSRG-PT2 code to a class
- [x] Decouple DWMS-DSRG-PT2 with FCI_MO
- [x] Fix reference energy of DWMS-AVG0 and DWMS-AVG1
- [x] Fix test cases


lcyyork:
@fevangelista No, the densities are the same as before. Otherwise the correlation energies will change. I only update the reference values for DWMS-AVG0 and DWMS-AVG1 where the MK vacuum is computed incorrectly. Now the MK vacuum is computed using the Reference class.

There is one caution though. Since we no longer rediagonalize the Hamiltonian after the semi-canonicalization, the CI vectors in FCI_MO class are in original basis. To make things consistent, I have to rotate semi-canonical integrals back to original for each root.

Thanks for the fast reviewing. I am actually thinking to make more changes as there is really no need to do a SA-DSRG-PT2 to just get the initial guess for DWMS-1. Plus the NH3 case is still giving me a hard time. The simpler idea is to use the previous DWMS CI vectors as a initial guess hoping that an EOM-like approach gives correct ordering.

PR #51, opened 2018-01-06
CASSCF semicanonical

lcyyork:
Make FORTE CASSCF output semicanonical orbitals when it's finished.


lcyyork:
Sorry that I forgot to pull before push. Those DWMS pushes are irrelevant. The only push is **bec10c7**.

jeffschriber:
Looks good. Is the SEMI_CANONICAL option always true whenever DSRG is called?

lcyyork:
SEMI_CANONICAL is set to true by default. Since only one job can be computed at a time, a way to use Kevin's CASSCF as DSRG reference is to first run a CASSCF job and pass its wavefunction to the subsequent DSRG computation.

I add this for a similar reason to [this psi4 issue](https://github.com/psi4/psi4/issues/740). Otherwise there will be some arbitrariness for any subsequent jobs that use this code if any orbitals are frozen.

PR #54, opened 2018-02-05
ACI batched screening

jeffschriber:
Creates new batching algorithm for screening excited determinants.
There are also some leftover commits from when pushing to main was allowed.

- [x] Implement threaded batching algorithm
    - [X] Working implementation
    - [X] Optimize (detatch) batch process
    - [X] Implement priorities for batch sizing
- [x] Add tests
- [x] Ready to go


src/aci/aci_build_F.cc
```diff
@@ -864,4 +862,284 @@ void AdaptiveCI::get_core_excited_determinants(SharedMatrix evecs, DeterminantHa
...
+            for (auto& ii : noalpha[h]) {
+                for (auto& aa : nvalpha[h]) {
+                    new_det = det;
+                    new_det.set_alfa_bit(ii, false);
+                    new_det.set_alfa_bit(aa, true);
+                    size_t hash_val = Determinant::Hash()(new_det);
+                    if ( (hash_val % nbin) == bin) {
+                        double HIJ = fci_ints_->slater_rules_single_alpha(det, ii, aa) * c_I;
+                        if ((std::fabs(HIJ) >= screen_thresh_)) {
+                            if (A_b.count(new_det) == 0) {
```
fevangelista: This question is for @jeffschriber  and @tyzhang1993. This operation requires two lookups of the hash table. Is there a way we can get this done in one go? Perhaps something like http://en.cppreference.com/w/cpp/container/map/emplace

tests/methods/aci-18/input.dat
```diff
@@ -0,0 +1,66 @@
...
+  aci_select_type            aimed_energy	 
+  aci_enforce_spin_complete  true
+  aci_add_aimed_degenerate false
+  aci_project_out_spin_contaminants false
+  diag_algorithm full
+  aci_batched_screening true
+  aci_nbatch 2
+#  aci_scale_sigma 0.0
+  active_guess_size 100
+}
```
fevangelista: - [ ] Can we cleanup this input so that inputs are aligned and unused keywords removed?
- [ ] Can we add another test so we can check for more than 2 bins and/or multiple threads?

tests/methods/aci-18/output.ref
```diff
@@ -0,0 +1,1147 @@
...
+  Time spent building a_list   0.000623 s
+  Time spent building b_list   0.000625 s
+  Time spent building aa_list  0.000898 s
+  Time spent building bb_list  0.000890 s
+  Time spent building ab_list  0.002240 s
+  Time spent diagonalizing H:   0.007410 s
+
+    P-space  CI Energy Root   0        = -229.358528279106 Eh =   0.0000 eV
+
+  Using batched Q_space algorithm
```
fevangelista: It would be nice to improve the formatting here. Is there a way to print this out as a table?

src/aci/aci_build_F.cc
```diff
@@ -864,4 +862,284 @@ void AdaptiveCI::get_core_excited_determinants(SharedMatrix evecs, DeterminantHa
...
+                    if( sp < r) continue;
+
+                    // Generate aa excitations
+                    for (int i = 0, max_i = noalpha[p].size(); i < max_i; ++i) {
+                        int ii = noalpha[p][i];
+                        for (int j = (p == q ? i + 1 : 0), maxj = noalpha[q].size(); j < maxj; ++j) {
+                            int jj = noalpha[q][j];
+                            for (int a = 0, max_a = nvalpha[r].size(); a < max_a; ++a) {
+                                int aa = nvalpha[r][a];
+                                for (int b = (r == sp ? a + 1 : 0), maxb = nvalpha[sp].size(); b < maxb; ++b) {
```
fevangelista: Here all the calls to `size()` can probably be moved up in the loop structure to improve efficiency:

- [ ] `max_i = noalpha[p].size()`
- [ ] `max_a = nvalpha[r].size()`
- [ ] `maxb = nvalpha[sp].size()`

This also applies to several loops below.

src/aci/aci_build_F.cc
```diff
@@ -864,4 +862,284 @@ void AdaptiveCI::get_core_excited_determinants(SharedMatrix evecs, DeterminantHa
...
+                                    int bb = nvalpha[sp][b];
```
fevangelista: Here it might be convenient to have a reference up in the loop structure `const auto& nvalpha_sp = nvalpha[sp]` so that you can call `nvalpha_sp[b]` instead. I am not sure it will be faster, but it can avoid doing a double lookup. This might be the most critical point, but the same comment applies to `ii`, `jj`, and `aa`.

src/aci/aci_build_F.cc
jeffschriber: Yes, it seems that unordered_map initializes values to 0.0, so we can get away with a single A_b[new_det] += HIJ call without any other lookups or logic.

PR #56, opened 2018-02-15
temporary fix the semicanonical problem when doing DF

lcyyork:
As stated in the title.


src/semi_canonicalize.cc
```diff
@@ -75,10 +75,22 @@ void SemiCanonical::startup() {
...
 
+    Ua_t_.iterate([&](const std::vector<size_t>& i, double& value) {
+        if (i[0] == i[1])
+            value = 1.0;
+    });
+    Ub_t_.iterate([&](const std::vector<size_t>& i, double& value) {
+        if (i[0] == i[1])
+            value = 1.0;
+    });
+
```
fevangelista: Are these loops doing something different than what is accomplished by the function `->identity();`?

PR #57, opened 2018-02-18
Semicanonical

lcyyork:
Sorry for being back and forth.

I found there is no guarantee that Ua (or Ub) will be identity when calling `semicanonicalize` unless the function call is directly after the `SemiCanonical` object is created. This is important in MRDSRG iterative relaxation when Fock matrix is block diagonal already at some point but Ua is from last iteration and not identity.

I guess at the end of the day the error will be very small, but this is an inconsistency. Please see the comments of commit `c838668`.

### DIFF
--- a/src/semi_canonicalize.cc
+++ b/src/semi_canonicalize.cc
@@ -172,11 +172,9 @@ void SemiCanonical::semicanonicalize(Reference& reference, const int& max_rdm_le
     // Check Fock matrix
     bool semi = check_fock_matrix();
 
-    // Set U to identity
-    set_U_to_identity();
-
     if (semi) {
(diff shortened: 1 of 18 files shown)


src/semi_canonicalize.cc
```diff
@@ -297,6 +292,27 @@ bool SemiCanonical::check_fock_matrix() {
     return semi;
 }
 
+void SemiCanonical::set_U_to_identity() {
+    Ua_->identity();
+    Ub_->identity();
+
+    Ua_t_.iterate([&](const std::vector<size_t>& i, double& value) {
+        if (i[0] == i[1]) {
```
fevangelista: Can you please shorten this and the next lines to:
`value = (i[0] == i[1]) ? 1.0 : 0.0;`

PR #58, opened 2018-02-25
Added variable RELAX_E_CONVERGENCE to control relaxation convergence.

tyzhang1993:
MRDSRG relaxation convergence was controlled by e_convergence. This PR separates the convergence criteria so that relaxation process can converge at a larger energy difference where the convergence of HF and CASSCF is not affected.

- [ ] Ready to go.


tyzhang1993:
Hi Francesco, thank you for the merge. However, this PR is not ready to go, I haven’t actually tested it. I just started the PR for related work. I am not sure if it actually works.

PR #61, opened 2018-03-14
Code optimization under UI64Determinant compilation

jeffschriber:
This PR is contains numerous optimizations when using the 64 bit representation of determinants, including:

- Batched ACI screening
- Direct RDM solver
- ACI-DSRG-MRPT3 w/ reference relaxation
- Direct computation of S2 eigenvalue

I think this branch should be merged in to master now, but with the branch kept alive for further optimization to the screening algorithm.

- [X] Ready to go


src/aci/aci.h
```diff
@@ -332,15 +332,33 @@ class AdaptiveCI : public Wavefunction {
...
     /// Builds excited determinants in batch
     det_hash<double> get_bin_F_space(int bin, int nbin, SharedMatrix evecs,
-                                    DeterminantHashVec& P_space); 
+                                     DeterminantHashVec& P_space);
+    det_hash<double> get_bin_F_space2(int bin, int nbin,double E0, SharedMatrix evecs,
+                                      DeterminantHashVec& P_space);
+    std::pair<std::vector<std::vector<std::pair<Determinant, double>>>, std::vector<size_t>>
+    get_bin_F_space3(int bin, int nbin, SharedMatrix evecs, DeterminantHashVec& P_space);
+    std::vector<std::pair<Determinant, double>>
+    get_bin_F_spacev(int bin, int nbin, SharedMatrix evecs, DeterminantHashVec& P_space);
```
tyzhang1993: What do these '1','2','3','v' mean after `get_excited_determinants_batch` and `get_bin_F_space`?
Can be more readable if you use more self-explanatory function names or add comments.

fevangelista: Good point. @jeffschriber, could you fix this? I am responsible for `get_excited_determinants_batch3` I think. Can you change this to `get_excited_determinants_batch_sort`?

src/ci_rdm/ci_rdms.h
```diff
@@ -244,6 +258,12 @@ class CI_RDMS {
 
     // Generate three-particle map
     void get_three_map();
+
+    // Fill 3rdm
+    void fill_3rdm( std::vector<double>& tprdm, double value, int p, int q, int r, int s, int t, int u , bool half = false);
+    void make_ab(std::vector<double>& tprdm_aab, std::vector<double>& tprdm_abb );
+    void make_ab(SortedStringList_UI64 a_sorted_string_list_,const  std::vector<UI64Determinant::bit_t>& sorted_astr,const std::vector<UI64Determinant>& sorted_a_dets, 
+                std::vector<double>& tprdm_ab, std::vector<double>& tprdm_aab,std::vector<double>& tprdm_abb);
```
fevangelista: Comment these functions a bit more argument. It would be nice to leave a note on storage format of RDMs somewhere in this code.

src/sparse_ci/ui64_determinant.cc
```diff
@@ -287,6 +287,70 @@ double UI64Determinant::destroy_beta_bit(int n) {
...
+    for( int h = 0; h < nirrep; ++h ){
+        for( int p = 0; p < act_mo[h]; ++p ){
+            if (not get_beta_bit(abs)) {
+                occ[h].push_back(abs);
+            }
+            abs++;
+        }
+    }
+    return occ;
+}
```
fevangelista: All of these look very inefficient due to allocation of multiple vectors. Are they called for each determinant? It might make sense to pass a `std::vector<std::vector<int>>` by reference and fill it up.

src/sparse_ci/ui64_determinant.h
```diff
@@ -33,14 +33,15 @@
 #include <cstdint>
 #include <memory>
 #include <vector>
+#include <bitset>
```
fevangelista: Why is this needed?

src/sparse_ci/ui64_determinant.h
```diff
@@ -136,7 +148,18 @@ class UI64Determinant {
...
+         //   return ((bs.a_ * 31) + bs.b_) ;
+          //  return ((bs.a_ * 31) + bs.b_) % 7561 ;
+          //  return ((bs.a_ * 31) + bs.b_) % 999331;
+   //       return ((std::hash<bit_t>()(bs.a_) * 13466917) + std::hash<bit_t>()(bs.b_)) % 1405695061;
+          return ((bs.a_ * 13466917) + bs.b_) % 1405695061;
+           // return ((bs.a_ * 31) + bs.b_) % 1405695061;
+           // std::bitset<128> big = bs.a_;
+           // big = (big << 64) | static_cast<std::bitset<128>>(bs.b_);
+           // return std::hash<std::bitset<128>>()(big);
         }
```
fevangelista: Please cleanup this code.

src/sparse_ci/ui64_determinant.h
```diff
@@ -152,6 +175,7 @@ uint64_t lowest_one_idx(uint64_t x);
 uint64_t clear_lowest_one(uint64_t x);
 
 double ui64_slater_sign(uint64_t x, int m, int n);
+double ui64_slater_sign(uint64_t x, int m);
```
fevangelista: This is in part my fault. Can you please comment explaining the difference between these two functions?

src/sparse_ci/ui64_determinant.cc
jeffschriber: Yeah, this is called for every reference determinant. I'll work on a better implementation of these functions (norb isn't even used?). Passing the object pre-allocated does make sense.

PR #65, opened 2018-04-10
Print the reference energy immediately after recomputing.

tyzhang1993:
This PR prints the recomputed reference energy before the DSRG iterations.
It enables estimation of results during computation when a job takes a long time.
- [x] Ready to go.


lcyyork:
I would probably just say "Updated reference energy E(0) = ...".

lcyyork:
It seems travis yml file is failing.

tyzhang1993:
@lcyyork Thanks! Changed the wording.
The travis problem seems to be problem with psi4's endorsed plugin.

fevangelista:
Yes, I am working on making Forte and Psi4 work together after the update of pybind11.

PR #71, opened 2018-06-26
Fix DMRG

jeffschriber:
CheMPS2 needs to be updated for the current Forte code. Not sure how much work this is.

- [ ] Ready to go


jeffschriber:
Sometimes this happens where we get a Travis compilation error which is largely erroneous. That aside, is this PR ready to merge?

PR #72, opened 2018-07-05
NIVO approximation for standard MR-DSRG

tyzhang1993:
We have implemented NIVO approximation for DF-SEQ-LDSRG(2). However, we have not tested the performance on regular DF-LDSRG(2).
This PR implements NIVO approximation for standard DF-LDSRG(2).

- [x] Ready to go.


fevangelista:
@tyzhang1993: Could you please add a section to the manual explaining the new options? I know that a DSRG section does not exist, so this is an opportunity to start one. Just a simple one with only the options for the DSRG would just be fine.

PR #73, opened 2018-07-27
Det cleanup

fevangelista:
This PR collects changes to the determinant and integral classes and documentation.

### Summary of changes
- [x] Use popcnt for architectures with SSE4.2 support
- [x] Refactor the integral class and eliminate redundant code
- [x] Document the integral class
- [x] Add ability to generate options list with defaults/descriptions
- [x] Add test cases for most integral types
- [x] Introduce new printing functions

### Status
- [x] Ready to go


src/integrals/integrals.cc
```diff
@@ -61,20 +62,26 @@ namespace forte {
...
+                                                   {Custom, "Custom"}};
+
 void set_INT_options(ForteOptions& foptions) {
     /*- The algorithm used to screen the determinant
      *  - CONVENTIONAL Conventional two-electron integrals
      *  - DF Density fitted two-electron integrals
      *  - CHOLESKY Cholesky decomposed two-electron integrals -*/
     foptions.add_str("INT_TYPE", "CONVENTIONAL",
-                     {"CONVENTIONAL", "DF", "CHOLESKY", "DISKDF", "DISTDF", "ALL", "OWNINTEGRALS"},
+    {"CONVENTIONAL", "DF", "CHOLESKY", "DISKDF", "DISTDF", "OWNINTEGRALS", "CUSTOM"},
```
lcyyork: What is the difference between OWNINTEGRALS and CUSTOM?

tests/methods/integrals-1/input.dat
```diff
@@ -0,0 +1,34 @@
...
+  job_type fci
+  frozen_docc      [2 ,0 ,0 ,0]
+  restricted_docc  [2 ,0 ,0 ,0]
+  active           [2 ,2 ,2 ,2]
+  e_convergence 12
+}
+
+energy('scf')
+compare_values(refscf, get_variable("CURRENT ENERGY"),10, "SCF energy") #TEST
+
```
lcyyork: Minor thing here. If the reference for forte is not set, a scf will run automatically. Thus, there is no need to run `energy('scf')`. Or it should be the following
```
Escf, wfn = energy('scf', return_wfn=True)
energy('forte', ref_wfn=wfn)
```
Since the documentation is limited, we should make our test cases the best guide.

src/integrals/integrals.cc
fevangelista: OwnIntegrals completely skips the generation of two-electron integrals. I am not sure if they are ever used in practice. Custom, which is currently not working and probably should be called FCIDUMP or something similar, is for reading a set of integrals from disk, potentially from other programs. Both need some work, and so it might make sense to remove these options from here. Does any of you ever use OwnIntegrals?

docs/sphinx/integrals.rst
```diff
@@ -0,0 +1,111 @@
...
+.. index::
+   single: Integrals
+   pair: Integrals; theory
+
+.. _`sec:integrals`:
+
+Two-electron Integral Types
+===========================
+
+.. codeauthor:: Kevin P. Hannoni, Chenyang Li, Francesco A. Evangelista
```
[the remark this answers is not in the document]
fevangelista: Thanks for catching this!

fevangelista:
We should continue our discussion of the frozen core energy in another PR.

PR #74, opened 2018-07-27
Test ambit

tyzhang1993:
This RP adds a test on ambit.

- [x] Ready to go.


src/test_ambit.cc
```diff
@@ -0,0 +1,1999 @@
...
+        }
+
+        if (report_result == kException)
+            outfile->Printf("\n    Unexpected: %s", exception.c_str());
+        if (report_result != kPass)
+            success = false;
+    }
+    outfile->Printf("\n %s", ::std::string(83, '-').c_str());
+    outfile->Printf("\n Tests: %s\n", success ? "All passed" : "Some failed");
+
```
lcyyork: Is this file basically copied from ambit?

tyzhang1993:
@lcyyork Sure, I can generate the output.ref files.
This file is pretty much a copy of the `test_operator.cc` in ambit. I will add this info in the file.

src/test_ambit.cc
```diff
@@ -0,0 +1,1672 @@
...
+}
+
+bool test_ambit() {
+    srand(time(nullptr));
+
+    auto test_functions = {
+        //            Expectation,  test function,  User friendly description
+        ::std::make_tuple(kPass, test_wrapper,
+                          "8 permutations of C(\"ij\") += A(\"ik\") * B(\"kj\")"),
+        ::std::make_tuple(kPass, test_C_equal_2_A, "C(\"ij\") = 2.0 * A(\"ij\")"),
```
fevangelista: This looks a bit funny, why is the "::" prefix needed?

src/test_ambit.cc
fevangelista: I think so.

src/test_ambit.cc
tyzhang1993: Oh, please look at the current Travis failed check as I removed "::". I do not fully understand but this is what I had to do in https://github.com/evangelistalab/forte/pull/74/commits/3328564f2da0480b861d8a6ff2a6f60b1c2a047c

PR #77, opened 2018-08-07
Corrected performance test.

tyzhang1993:
This PR corrected the performance tests.

- [x] Ready to go.

### DIFF
--- a/tests/performance/aci-1/input.dat
+++ b/tests/performance/aci-1/input.dat
@@ -38,10 +38,10 @@ set forte {
 
 energy('scf')
 
-#compare_values(refscf, get_variable("CURRENT ENERGY"),9,"SCF energy")
+compare_values(refscf, get_variable("CURRENT ENERGY"),9,"SCF energy")
 
 energy('forte')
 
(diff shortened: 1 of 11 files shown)


tyzhang1993:
@fevangelista (i) I think that was because of the basis set change. These tests are 3 years old and pretty outdated.
(ii) I found that originally in these test cases, some of them have result check, some of them do not. In order to have a clean output, I commented out all the result checks for I think the correctness of algorithm should be checked in `test/method` rather than in the performance test. I can add them back if you think that is better idea.

fevangelista:
I think we should reintroduce the tests to make sure the timing is for a procedure that ran correctly.

tyzhang1993:
Good point, I will.

jeffschriber:
Code looks good, but could you pull/rebase master to make sure travis compiles and runs?

tyzhang1993:
@jeffschriber I actually rebased master. Travis fails because of psi4 compiling error rather than forte problem.

jeffschriber:
Oh I understand now. Yes we can merge once psi is working.

PR #80, opened 2018-09-05
Modify .travis.yml to test both determinant implementations.

tyzhang1993:
This PR modified .travis.yml to test two different implementations of determinants by passing environment variables through `-DMAX_DET_ORB=<num_str_bits>`.

# Usage ##
If `MAX_DET_ORB<=64`, ui64_determinant is used. Otherwise, stl_bitset_determinant is used with the number of string bits specified by `-DMAX_DET_ORB`.

- [x] Ready to go.


tyzhang1993:
@jeffschriber What's the best way to figure out which version is running?

jeffschriber:
I'd do `sizeof(Determinant)` to see if it is 16 or 32

tyzhang1993:
@jeffschriber  I mean if there is a easy way to tell if the two travis tests are running different implementations of Determinant. Currently there is no difference in there outputs.

jeffschriber:
The output should print the size of the determinant being used just under the Forte banner

tyzhang1993:
Thank @jeffschriber ! I successfully printed the size of det at the end of test. However, both of the travis tests returns det size 16. I compile with `-DSMALL_BITSET=${SMALL_BITSET}`, where SMALL_BITSET is either `true` or `false`. Is there anything wrong with my compilation command?

src/aci/aci_build_F.cc
```diff
@@ -1182,7 +1182,7 @@ AdaptiveCI::get_excited_determinants_batch_vecsort(SharedMatrix evecs, SharedVec
                 double& V = pair.second;
                 if (V != 0.0) {
                     auto& det = pair.first;
-                    double delta = fci_ints_->energy_bit(det) - E0;
```
fevangelista: Thanks for cleaning up this. I think it is a good solution to remove `energy_bit` and to have only one function.

fevangelista: Ok, I just saw the code below and now I see that the new `energy(det)` is the efficient bit version for 64 bits and the standard one for larger orbital spaces.

tyzhang1993:
@fevangelista In Travis CI, the two different environments compiles two different versions of the code. So both 64 and 128 implementations are tested for all test cases in Travis.

jeffschriber:
I think there is one other issue we should clear up before merging. The direct RDM builder in ci_rdms_dynamic.cc uses explicit calls to UI64Determinant throughout, and still compiles and runs correctly when Determinant is defined with more than 64 bits. This code makes heavy use of `clear_lowest_one()`, `lowest_one_idx()`, and the SortedStringList_UI64 class in general. Should this code only be available in the 64 bit compilation, or should it be generalized to work with all compilations?

jeffschriber:
Related to the above comment, this means that `aci-dsrg-mrpt2-3`, `aci-dsrg-mrpt2-4`, and `aci-dsrg-mrpt2-5` will fail if they are modified to use more than 64 spin orbitals

fevangelista:
@tyzhang1993: is there a test case that validates the 128 bit implementation (i.e., uses more than 64 orbitals?). Let me know when you add it and I will come back to approve.

tyzhang1993:
@fevangelista Yes, but it passes on any other computer but on travis. It is located at https://github.com/evangelistalab/forte/blob/travis_det_size/tests/large_det/run_forte_tests_travis.py
Turned on in travis by https://github.com/evangelistalab/forte/blob/travis_det_size/.travis.yml#L191
I have not solved the problem had no clue how.

tyzhang1993:
@fevangelista I have added it, you can approve this PR.

PR #83, opened 2018-10-23
Add function to print final wave function to a file

jeffschriber:
Very small pr, but something that needed to be cleaned up.
The goal is to add a function and corresponding option to print an ACI wave function to a file. The function will create a .txt file with the expansion coefficient and corresponding occupation vector printed on each line. 

- [X] Add the printing function

- [x] Ready to go

### DIFF
--- a/src/aci/aci.cc
+++ b/src/aci/aci.cc
@@ -40,14 +40,6 @@ using namespace psi;
 namespace psi {
 namespace forte {
 
-//#ifdef _OPENMP
-//#include <omp.h>
-//#else
-//#define omp_get_max_threads() 1
-//#define omp_get_thread_num() 0
-//#define omp_get_num_threads() 1
-//#endif
-
 void set_ACI_options(ForteOptions& foptions) {
     /* Convergence Threshold -*/
     foptions.add_double("ACI_CONVERGENCE", 1e-9, "ACI Convergence threshold");
--- a/src/aci/aci.cc
+++ b/src/aci/aci.cc
@@ -1460,10 +1460,10 @@ std::vector<std::pair<double, double>> AdaptiveCI::compute_spin(DeterminantHashV
 void AdaptiveCI::wfn_to_file(DeterminantHashVec& det_space, SharedMatrix evecs, int root) {
 
     std::ofstream final_wfn;
-    final_wfn.open("final_wfn_"+ std::to_string(root) +  ".txt");
+    final_wfn.open("aci_final_wfn_"+ std::to_string(root) +  ".txt");
     const det_hashvec& detmap = det_space.wfn_hash();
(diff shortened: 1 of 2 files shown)


src/aci/aci.cc
```diff
@@ -1461,90 +1457,17 @@ std::vector<std::pair<double, double>> AdaptiveCI::compute_spin(DeterminantHashV
...
-/*
-void AdaptiveCI::wfn_analyzer(DeterminantMap& det_space, SharedMatrix evecs, int nroot) {
+void AdaptiveCI::wfn_to_file(DeterminantHashVec& det_space, SharedMatrix evecs, int root) {
 
-    std::vector<bool> occ(2 * nact_, 0);
-    std::vector<std::tuple<double, int, int>> labeled_orb_en = sym_labeled_orbitals("RHF");
-    for (int i = 0; i < nalpha_; ++i) {
-        occ[std::get<2>(labeled_orb_en[i])] = 1;
+    std::ofstream final_wfn;
+    final_wfn.open("final_wfn_"+ std::to_string(root) +  ".txt");
```
fevangelista: Could we make this "aci_final_wfn" to make sure it does not get confused with other types?

src/aci/aci.cc
```diff
@@ -1461,90 +1457,17 @@ std::vector<std::pair<double, double>> AdaptiveCI::compute_spin(DeterminantHashV
...
+    const det_hashvec& detmap = det_space.wfn_hash();
+    for (size_t I = 0, maxI = detmap.size(); I < maxI; ++I) {
+        final_wfn << std::setw(18) << std::setprecision(12) << evecs->get(I,root) << "\t" << detmap[I].str(nact_).c_str() << std::endl;
```
fevangelista: The best thing to do here is to print the coefficients in scientific notation with a fixed number of digits (9-12).

src/aci/aci.cc
jeffschriber: Yes this makes sense

tyzhang1993:
Could you please give an output example in the description of the PR or explain the output format? Thanks!

jeffschriber:
I think integral printing could be useful, but should be in a different PR. Francesco also mentioned that accessing the integrals in Python should be done by calling a function rather than reading from disk.

tyzhang1993:
@jeffschriber I am not sure about this. Definitely, integrals can get by directly call psi4 API functions. However, sometimes orbitals are actually rotated in the printed determinants which do not match the integrals directly get from psi4 API. Consistency is my actual point.

fevangelista:
@tyzhang1993: I mentioned to @jeffschriber that I would prefer to have a different mechanism to print out the integrals (e.g. FCIDUMP) and that it would be great to combine it with the ability to read integrals in. So we could do both exporting to python and restarting forte computations in one go.

PR #84, opened 2018-10-25
Fix test cases

jeffschriber:
Changes to physical constants in psi have broken some of our more stringent test cases. The changes in the energy are very small, on the order of ~1e-8. This PR fixes all broken tests so that the master branch will pass on travis.

This PR requires a version of psi no later than Sept 29 (ce75f8174f01538063dde07d5900bed967fc8dbd). 

The list of changed constants/masses is quite extensive, but it can be found in the initial PR from Psi at 
https://github.com/psi4/psi4/pull/1213

- [x] Ready to go


tyzhang1993:
Could you please put in this PR exactly which physical constants is changed and what's the change? Thanks!

PR #86, opened 2018-11-07
Deboost Forte and enable C++17

fevangelista:
**Description**
This PR removes boost from Forte, transitions the CMake build system to use C++17, and applies clang-tidy to use nullptr instead of 0 or NULL

**Changes relevant to users**
- Needs a compiler compatible with C++17
- Requires a pre-compiled boost library


src/mrdsrg-spin-integrated/mrdsrg_smart_s.cc
```diff
@@ -142,7 +142,7 @@ double MRDSRG::smart_s_davg_min_delta1() {
     }
     double davg_sum = std::accumulate(davg.begin(), davg.end(), 0.0);
     std::transform(davg.begin(), davg.end(), davg.begin(),
-                   std::bind1st(std::multiplies<double>(), 1.0 / davg_sum));
+                   [&](double x) { return x / davg_sum; });
```
fevangelista: @lcyyork : can you check that this is the correct translation of the `bind1st` call here? `bind1st` is not supported in C++17 and a lambda should be the simplest solution.

lcyyork: @fevangelista Thanks for the modification. This should be good.

fevangelista: Great. If you have chance please review and comment/approve

lcyyork: I do not have any problem so far.

PR #104, opened 2018-12-16
Move Forte out of namespace psi

fevangelista:
## Description
Moves Forte out of namespace psi

## Checklist
- [x] Ready to go!


fevangelista:
@lcyyork, @jeffschriber, @tyzhang1993: This is an experimental PR. Please take a look and let me know what you think about it. I don't want it to be too burdensome to work with `psi::` tags everywhere in the code.

src/fci/fci.cc
```diff
@@ -35,7 +35,8 @@
 
 #include "fci.h"
 
-namespace psi {
+using namespace psi;
```
jeffschriber: Keep namespace here?

src/fci/fci_vector.cc
```diff
@@ -35,11 +35,12 @@
 #include "helpers/mo_space_info.h"
 #include "fci_vector.h"
 
-namespace psi {
+using namespace psi;
```
jeffschriber: Namespace declared here as well

src/fci/fci.cc
fevangelista: Here I am using the namespace. This allows me to avoid using `psi::outfile->` everywhere. To remove the name space completely we would have to isolate Forte from psi4.

fevangelista:
We could wrap a call to `outfile->Printf()` with a Forte output object. This could be combined in an object that also manages warnings if want. Let me play a bit around.

PR #110, opened 2019-01-08
Multistate active space solver

fevangelista:
## Description
This PR introduces a multi-state active space solver class (MSActiveSpaceSolver) and standardizes the interface of all the methods derived from ActiveSpaceSolver.

## User Notes
- [ ] Features added

## Checklist
- [ ] Documented source code
- [ ] Ready to go!

## Failing test cases (update on 1/15 5pm)
+ actv-dsrg:
  - [ ] actv-dsrg-4-C4H6-triplet
  - [ ] actv-dsrg-ipea-2
+ dsrg-mrpt2:
  - [x] dsrg-mrpt2-8-sa (the sa-sub should not run sa-full procedure)
  - [ ] dsrg-mrpt2-9-xms
  - [x] dsrg-mrpt2-10-CO
+ dsrg-mrpt3:
  - [x] dsrg-mrpt3-7-CO
+ dwms-dsrgpt:
  - [ ] dwms-dsrgpt-1-sa
  - [ ] dwms-dsrgpt-2-sa
  - [ ] dwms-dsrgpt-4-sa
  - [ ] dwms-dsrgpt-5-ms
  - [ ] dwms-dsrgpt-8-separated
  - [ ] dwms-dsrgpt-3-sa
  - [ ] dwms-dsrgpt-6-ms
  - [ ] dwms-dsrgpt-7-ms
  - [ ] dwms-dsrgpt-9-sa-C6H6


fevangelista:
@lcyyork, @tyzhang1993, @jeffschriber : I am wondering if perhaps we should change the name of Reference -> RDMS or something like that. With the MS formulations of many of our methods, we will need transition RDMS and so it might make sense to have an object that can store diagonal and off-diagonal RDMS. Also, the ActiveSpaceSolver should have a function:

```c++
RDM rdms(size_t l_reference, size_t r_reference) const;
```

that returns a generic transition density matrix.

lcyyork:
I am decoupling fcimo and mcsrgpt2mo. Would you postpone the name change for now? It is true that for multistate calculations I just use Reference as a transition or normal density container. Active space solver should have the functionality of compute transition density in fcimo. However, I think it should be in active space solver itself because it should be able to handle states of different symmetry. Also, sometimes there is no need to recompute the densities, like in the dw-dsrg-pt2, it is more convenient to dump to disk, like done in fcimo.

tyzhang1993:
I agree. Currently, the Reference class is essentially an RDMS class. Active space solver and RDMs are two different things, they should definitely be different classes, but Active space solver can compute and hold RDMs objects.

src/base_classes/active_space_solver.cc
```diff
@@ -26,72 +26,423 @@
...
+        auto energies = method->energies();
+        for (size_t r = 0; r < nroot; r++) {
+            energy += energies[r] * weights[r];
+        }
+        method_vec_.push_back(method);
+    }
+    return energy;
+}
+
+Reference ActiveSpaceSolver::get_reference(int root) {
```
fevangelista: I suppose we need this for back compatibility but we can at least start to do a couple of things to make the purpose of this function clearer:

- wrap the state-averaged procedure in its own function that is called on line 91.
- put the code that transforms the RDM <-> cumulant in a separate function in the reference.cc file. This should go in a free function and not in the class.
- the input of this functions should be generalized to get a pair (state symmetry, root). But right now it is more important to pass the root variable down to the other methods.

src/base_classes/active_space_solver.cc
```diff
@@ -26,72 +26,423 @@
...
+
+    Reference ref;
+
+    // For single state
+    if( method_vec_.size() == 1 ){
+        ref =  method_vec_[0]->get_reference();
```
fevangelista:
```suggestion
       ref = method_vec_[0]->get_reference(root);
```

fevangelista: Also, can you pleas apply clang-format before commits? I think there are some inconsistencies in the formatting.

src/base_classes/active_space_solver.cc
jeffschriber: I have some un-pushed commits that do much of this work. Currently, the SA references are built properly using the ActiveSpaceSolver interface (even for FCI_MO!), but the final computation of the energy/oscillator strengths does not work since each DSRG method handles that separately at the moment. The issue is that that part of the code needs knowledge of the determinants for each state. A quick fix is to implement a method->determinants() function, though generally I don't think we want this.

lcyyork:
@fevangelista I temporarily create a contractedCI class. I am not sure this is an ideal solution because it might only be used for the contracted variant of sa-dsrg. For dw-dsrg, the weights are changing and the weights are determined outside the ActiveSpaceSolver unless they are determined by the reference energy.

pymodule.py
```diff
@@ -37,21 +37,29 @@
...
     state_energies_list = active_space_solver.compute_energy()
 
-#    correlation_solver_type = options.get_str('CORRELATION_SOLVER')
-#    if correlation_solver_type != 'NONE':
+    correlation_solver_type = options.get_str('CORRELATION_SOLVER')
+    if correlation_solver_type != 'NONE':
+        reference = active_space_solver.reference()
+        semi = forte.SemiCanonical(mo_space_info, ints, options)
+        semi.semicanonicalize(reference, max_rdm_level)
+        Ua = semi.Ua_t()
```
lcyyork: @fevangelista Do you have any idea what is wrong here (`semi.Ua_t()`)? It keeps saying `TypeError: Unable to convert function return value to a Python type! The signature was
	(self: forte.forte.SemiCanonical) -> ambit::Tensor` on my computer.

fevangelista: Could the problem be that ambit::Tensor is not exported via pybind11?

fevangelista: Ambit wrappers are available for Python, we should try to add `import ambit`

fevangelista: Ok, it just need exposing `ambit::Tensor' via the one liner
```c++
    py::class_<ambit::Tensor>(m, "ambitTensor");
```

lcyyork: Yes, thanks for the push!

fevangelista: You're set to go. I just exposed `make_dynamic_correlation_solver`, but you will need to return an object before being able to use it.

lcyyork: I currently expose the master_dsrg and the procedure should work for all unrelaxed dsrg methods.

fevangelista:
Test `MRDSRG-PT2-4` is failing after the last commit.

lcyyork:
@jeffschriber Thanks for the change. Just about to do it.

fevangelista:
I was wondering: it looks to me that we will need the user to specify what type of calculation is running (SS, SA, MS, DWMS, ...). How about introducing an option that controls this part? Right now some of this info is deduced (?) and stored for convenience in the active space solver object. Should we redefine JOB_TYPE or introduce something new like CALC_TYPE?

fevangelista:
Can someone look into?
- [ ] CD-DSRG-MRPT2-7-SA

fevangelista:
Thanks @lcyyork. Next things to do:

- Remove is_multistate and other variables from `ActiveSpaceSolver`
- Restructure the python driver to make decision based on CALC_TYPE.

fevangelista:
@lcyyork , @tyzhang1993 , @jeffschriber: Once the above are addressed, should we pull this PR in and then make a cleanup/review-PR? That way it would be easier to leave comments in the files.

tyzhang1993:
Yes, I think we can do that.

src/base_classes/active_space_solver.h
```diff
@@ -96,21 +96,6 @@ class ActiveSpaceSolver {
...
-    /// Is this ActiveSpaceSolver targets a multi-state computation
-    bool is_multi_state() { return state_specific_root_ < 0 ? true : false; }
-
-    const std::map<StateInfo, size_t>& get_state_list() const { return state_list_; }
-
-    const std::map<StateInfo, std::shared_ptr<ActiveSpaceMethod>>& get_method_map() const {
-        return method_map_;
-    }
-
-    const std::map<StateInfo, std::vector<double>>& state_energies_map() const {
```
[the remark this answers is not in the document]
lcyyork: I reintroduce it.

jeffschriber:
Any reason not to delete this branch?

PR #116, opened 2019-01-17
CI framework for selected CI methods.

tyzhang1993:
## Description
This PR creates the framework for selected CI methods: ACI, PCI. The ultimate goal is to build a superclass with CI workflow and expose the interfaces such as sigma builder, determinant space expander.

## New classes
- [x] `ExcitedStateSolver` a subclass of `ActiveSpaceMethod`, controlling the excited states computation.
- [x] `SelectedCIMethod` a superclass of ACI and PCI.
- [x] `PCISigmaVector` a subclass of `SigmaVector`, hides all the logic of sigma build (spawning). Enable integration of PCI and SparseCISolver.

## Major changes
- [x] Separated codes for excited states and ACI iterations in the original AdaptiveCI class.
- [x] Reenabled PCI excited state calculation.
- [x] Removed replicated PCI codes.

## Near Future work
- Remove `ex_alg` variable from `SelectedCIMethod::set_method_variables`
- Merge `AdaptiveCI_SCI` to `AdaptiveCI` and remove replicated codes.
- Code clean up.

## Status
- [x] Ready to go!


src/ci_ex_states/excited_state_solver.cc
```diff
@@ -0,0 +1,1131 @@
...
+    sparse_solver_->set_parallel(true);
+    sparse_solver_->set_force_diag(options->get_bool("FORCE_DIAG_METHOD"));
+    sparse_solver_->set_e_convergence(options->get_double("E_CONVERGENCE"));
+    sparse_solver_->set_maxiter_davidson(options->get_int("DL_MAXITER"));
+    sparse_solver_->set_spin_project(options->get_bool("ACI_PROJECT_OUT_SPIN_CONTAMINANTS"));
+    sparse_solver_->set_spin_project_full(options->get_bool("ACI_PROJECT_OUT_SPIN_CONTAMINANTS"));
+    sparse_solver_->set_guess_dimension(options->get_int("DL_GUESS_SIZE"));
+    sparse_solver_->set_num_vecs(options->get_int("N_GUESS_VEC"));
+    sparse_solver_->set_sigma_method(options->get_str("SIGMA_BUILD_TYPE"));
+    sparse_solver_->set_max_memory(options->get_int("SIGMA_VECTOR_MAX_MEMORY"));
```
jeffschriber: We should think about changing the option names with ACI to SCI or something more general

src/ci_ex_states/excited_state_solver.cc
```diff
@@ -0,0 +1,1131 @@
...
+                                       std::shared_ptr<MOSpaceInfo> mo_space_info,
+                                       std::shared_ptr<ActiveSpaceIntegrals> as_ints,
+                                       std::unique_ptr<SelectedCIMethod> sci)
+    : ActiveSpaceMethod(state, nroot, mo_space_info, as_ints), sci_(std::move(sci)) {
+    nact_ = mo_space_info_->size("ACTIVE");
+}
+
+void ExcitedStateSolver::set_options(std::shared_ptr<ForteOptions> options) {
+    // TODO: This shouldn't come from options
+    root_ = options->get_int("ROOT");
```
jeffschriber: Can't we get this from ActiveSpaceSolver?

src/ci_ex_states/excited_state_solver.cc
```diff
@@ -0,0 +1,1131 @@
...
+                op_.op_lists(final_wfn_);
+                op_.tp_lists(final_wfn_);
+            } else {
+                op_.clear_op_s_lists();
+                op_.clear_tp_s_lists();
+                op_.build_strings(final_wfn_);
+                op_.op_s_lists(final_wfn_);
+                op_.tp_s_lists(final_wfn_);
+            }
+        }
```
jeffschriber: I think we can safely remove this add_singles functionality for now. It's never used, and if we eventually want to do this, I think a separate module makes more sense, i.e. a general uncontracted MRCI class.

src/ci_ex_states/excited_state_solver.cc
```diff
@@ -0,0 +1,1131 @@
...
+
+        if (max_rdm_level >= 3) {
+            psi::outfile->Printf("\n  Computing 3-list...    ");
+            local_timer l3;
+            op_.three_s_lists(final_wfn_);
+            psi::outfile->Printf(" done (%1.5f s)", l3.get());
+        }
+    }
+
+    CI_RDMS ci_rdms_(dets, fci_ints, PQ_evecs, root1, root2);
```
jeffschriber: This one is my fault from a long time ago, this variable name should be `ci_rdms` without the final underscore

src/ci_ex_states/excited_state_solver.cc
```diff
@@ -0,0 +1,1131 @@
...
+//        //        %.12f Eh = %8.4f eV",i,abs_energy + davidson[i],
+//        //                exc_energy + pc_hartree2ev * (davidson[i] -
+//        //                davidson[0]));
+//        //    	}
+//    }
+
+//    print_wfn(ref, op, final_evecs, nroot_);
+//    max_rdm_level_ = 1;
+//    compute_rdms(fci_ints, ref, op, final_evecs, 0, 0);
+//}
```
jeffschriber: Again, we can delete these commented lines. A lot of this code is in mrci.cc, where it should be

src/sci/aci_sci.cc
```diff
@@ -0,0 +1,1877 @@
...
+            stuck = true;
+        }
+    }
+    return stuck;
+}
+
+std::vector<std::pair<double, double>> AdaptiveCI_SCI::compute_spin(DeterminantHashVec& space,
+                                                                    WFNOperator& op,
+                                                                    psi::SharedMatrix evecs,
+                                                                    int nroot) {
```
jeffschriber: Wasn't this also a function in ExcitedStateSolver?

src/sci/aci_sci.cc
```diff
@@ -0,0 +1,1877 @@
...
+
+void AdaptiveCI_SCI::find_q_space() {
+    // Step 2. Find determinants in the Q space
+    local_timer build_space;
+    if (options_->get_bool("ACI_BATCHED_SCREENING")) {
+        find_q_space_batched(P_space_, PQ_space_, P_evals_, P_evecs_);
+    } else if (streamline_qspace_) {
+        default_find_q_space(P_space_, PQ_space_, P_evals_, P_evecs_);
+    } else {
+        find_q_space(P_space_, PQ_space_, num_ref_roots_, P_evals_, P_evecs_);
```
jeffschriber: This `find_q_space` should maybe have a different name to avoid confusion. Something like `find_q_space_multiroot`

src/ci_ex_states/excited_state_solver.h
```diff
@@ -0,0 +1,148 @@
...
+#else
+#define omp_get_max_threads() 1
+#define omp_get_thread_num() 0
+#define omp_get_num_threads() 1
+#endif
+
+namespace forte {
+class SelectedCIMethod;
+
+class ExcitedStateSolver : public ActiveSpaceMethod {
```
fevangelista: @tyzhang1993: I am not sure this class hierarchy is consistent with how we designed the new base classes. Other methods implement excited states (e.g. FCI) without creating an excited state class. Can we please discuss a different design where the various algorithms are implemented in a base class of SelectedCI and methods like ACI are derived directly from SelectedCI?

tyzhang1993: Yes, it is a good point.
I had this design in order to have functionality really separated. Unlike FCI where the excited states algorithm is straightforward, the excited states algorithm for selected CI methods is pretty independent and has many variations.
The role of excited state solver is to determine how to compute a state, it calls the selected CI method.
The role of selected CI method is performing a selection of important determinants and optimize the coefficients.
Excited state solver drives the computation of selected CI wave function. If they are in one class, I think it is strange to have a class drive itself.
For example, in the original ACI implementation, in order to "drive ACI" performing root_orthogonalized computation, the class variable nroot_ has to be set to 1 before solving the PQ space, then set back to original value. https://github.com/evangelistalab/forte/blob/e82b201ecd70cacb81fd0b7445cec6348b07ba8d/src/sci/aci.cc#L1663-L1672
In my opinion, different excited state algorithms are actually different subclasses of ExcitedStateSolver, in commit https://github.com/evangelistalab/forte/pull/116/commits/600b6544911dfaa7bee141293ec67b28200d2b84 , there are not only the superclass ExcitedStateSolver, but also GroundStateSolver, RootOrthoSolver etc. These excited states algorithms are very different from each other, having derived classes can make code more clear and avoid if statements. It is like our having different variations of sigmavector classes.
https://github.com/evangelistalab/forte/blob/9ee177bff82d50b6e1e1de9487f3151598afcc9b/src/ci_ex_states/excited_state_solver.cc#L169-L188
I didn't separate ExcitedStateSolver into these subclasses in this PR because I do not want the PR to change code too dramatic. In this PR I just move code around to create the structure. but I believe put excited state algorithms into subclasses of  ExcitedStateSolver should be done in the following PRs.
Because the ExcitedStateSolver and SelectedCIMethod are independent of each other and they both have many different variations, I think they should be different classes.

src/ci_ex_states/excited_state_solver.cc
tyzhang1993: Yes, but currently in order not to change the input files of aci and pci test cases dramatically, we can do this cleanup in new PRs.

jeffschriber: That makes sense

src/ci_ex_states/excited_state_solver.cc
tyzhang1993: This one I copied from ACI, I have not think about this variable in detail.
https://github.com/evangelistalab/forte/blob/9ee177bff82d50b6e1e1de9487f3151598afcc9b/src/sci/aci.cc#L127-L128
I think you are right, but it seems that currently we do not have a good clue where we set this variable.

PR #134, opened 2019-03-25
Exposed Determinant class.

tyzhang1993:
## Description
This PR exposes the Determinant Class.

## Exposed members
- [x] `Determinant::num_str_bits`
- [x] `Determinant::num_det_bits`
- [x] `Determinant()`
- [x] `Determinant(const std::vector<bool>& occupation_a, const std::vector<bool>& occupation_b)`
- [x] `Determinant::get_alfa_bits()`
- [x] `Determinant::get_beta_bits()`
- [x] `Determinant::set_alfa_bits()`
- [x] `Determinant::set_beta_bits()`
- [x] `Determinant::create_alfa_bits()`
- [x] `Determinant::create_beta_bits()`
- [x] `Determinant::destroy_alfa_bits()`
- [x] `Determinant::destroy_beta_bits()`
- [x] `Determinant::str()`
- [x] `ForteIntegrals::nmo()`
- [x] `ForteIntegrals::ncmo()`
- [x] `ActiveSpaceIntegrals(std::shared_ptr<ForteIntegrals> ints, std::shared_ptr<MOSpaceInfo> mospace_info)`
- [x] `ActiveSpaceIntegrals::slater_rules()`

## Example usage
```python
>>> import forte
>>> forte.Determinant.num_str_bits
64
>>> forte.Determinant.num_det_bits
128
>>> a = forte.Determinant([True,False,True],[False,True,True])
>>> a.get_alfa_bits()
5
>>> a.get_beta_bits()
6
>>> bin(a.get_beta_bits())[:1:-1]
'011'
>>> b = forte.Determinant()
>>> b.get_alfa_bits()
0
```

## User Notes
- [x] Corrected a bug in CMakeList to reenable `UI64Determinants`.

## Unsolved issue
- This PR currently only works with `UI64Determinants`, because the return type `std::bitset` of `STLBitsetDeterminant::get_alfa_bits()` is not cast successfully to python.
Error message: `TypeError: Unable to convert function return value to a Python type! The signature was
	(self: forte.forte.Determinant) -> std::__1::bitset<64ul>`

## Checklist
- [x] Ready to go!

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -192,6 +192,7 @@ if(MAX_DET_ORB)
     add_definitions(-DMAX_DET_ORB=${MAX_DET_ORB})
 else()
     add_definitions(-DMAX_DET_ORB=64)
+    add_definitions(-DSMALL_BITSET)
 endif()
 
 if(ENABLE_MPI)
--- a/src/python_api.cc
+++ b/src/python_api.cc
@@ -185,7 +185,9 @@ PYBIND11_MODULE(forte, m) {
 
     // export ForteIntegrals
     py::class_<ForteIntegrals, std::shared_ptr<ForteIntegrals>>(m, "ForteIntegrals")
-        .def("rotate_orbitals", &ForteIntegrals::rotate_orbitals);
+        .def("rotate_orbitals", &ForteIntegrals::rotate_orbitals)
+        .def("nmo", &ForteIntegrals::nmo)
+        .def("ncmo", &ForteIntegrals::ncmo);
 
     // export StateInfo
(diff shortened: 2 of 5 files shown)


fevangelista:
The problem with `std::__1::bitset<64ul>` is that there is no automatic way to convert this type to some Python type. I think we might be able to export the type in Python with pybind11, but then we have to expose the entire interface. If you expose the functions that return the occupation of an orbital, then we can still access the bits. Also, it would be nice to add the functions that expose the creation/annihilation of electrons.

tyzhang1993:
@fevangelista I have exposed the getters, setters, creators and destroyers.

PR #135, opened 2019-03-27
Access NO occupations in python

jeffschriber:
## Description
Allows access to NO occupations in python for analytical purposes
Any suggestions for exposure of any similar quantities are welcome here as well.

## Checklist

- [X] Create function to return labeled NO occupations
- [X] Expose function in python 
- [x] Documented source code
- [X] Checked for redundant headers
- [X] Checked for consistency in the formatting of the output file

- [x] Ready to go!

### DIFF
--- a/src/base_classes/rdms.cc
+++ b/src/base_classes/rdms.cc
@@ -320,6 +320,25 @@ double compute_Eref_from_rdms(RDMs& ref, std::shared_ptr<ForteIntegrals> ints,
     return E;
 }
 
+std::vector<std::vector<double>> get_rdm_data(RDMs& ref, size_t level){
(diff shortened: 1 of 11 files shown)


fevangelista:
Thanks Jeff. I realize that you have gone one step forward and also computed NOs. This is great, but can you add an option to skip NO formation? For the selector we need to report the occupation of each selected orbitals. Perhaps an extra argument (`compute_nos`, set to false by default) could be added to the function to toggle the NO computation (in this case we should probably change the function name to `get_occ`). Could you also add a function to return the 1-RDM as a std::vector in Python? That might be useful for computing the NOs directly on the Python side.

PR #140, opened 2019-04-18
Update full PT2 correction in sCI

jeffschriber:
## Description
This PR updates the class used for the full PT2 correction to sCI energies. It removes dependencies
on the psi::Wavefunction and other extraneous classes. 

### Checklist
- [X] Reintegrate MRPT2 class
- [X] Update MRPT2 code to handle multiple excited states
- [X] Documented source code
- [X] Checked for consistency in the formatting of the output file

- [x] Ready to go!

### DIFF
--- a/src/sci/sci.cc
+++ b/src/sci/sci.cc
@@ -64,6 +64,9 @@ double SelectedCIMethod::compute_energy() {
         prune_PQ_to_P();
     }
 
+    // Active space PT2 correction
+    full_mrpt2();
+
     // Post-iter process
     post_iter_process();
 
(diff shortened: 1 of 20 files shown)


src/sci/sci.cc
```diff
@@ -64,6 +64,9 @@ double SelectedCIMethod::compute_energy() {
         prune_PQ_to_P();
     }
 
+    // Active space PT2 correction
+    full_mrpt2();
```
tyzhang1993: I personally think `full_mrpt2()` should be called in `post_iter_process()`. This may not be common to all `sci` algorithms.

jeffschriber: I like that change, will modify asap

PR #144, opened 2019-04-20
Fix AVAS when SIGMA = 1.0

### LINKED ISSUE #143: AVAS-1 fails
I noticed some odd behavior with the avas-1 test. If I run it ten times or so, about a quarter of those runs fail due to the incorrect number of active orbitals being chosen. Any ideas @henankf223 or @fevangelista?

fevangelista:
## Description
Avoid problems with roundoff errors when AVAS_SIGMA = 1.0. To solve the problem I check that the sum of singular values `sum_i s_i <= avas_sigma + 1.0e-9`. So when `avas_sigma = 1.0` all orbitals are included. To avoid including orbitals with small singular value, I make sure that only orbitals with `s_i > 1.0e-6` are included in the list of actives. 

## Checklist
- [ ] Resolves #143
- [ ] Ready to go!

### DIFF
--- a/src/orbital-helpers/avas.cc
+++ b/src/orbital-helpers/avas.cc
@@ -210,8 +210,10 @@ void make_avas(psi::SharedWavefunction ref_wfn, psi::Options& options, psi::Shar
                 }
             }
         } else {
+            // tollerance on the sum of singular values (for border cases, e.g. sigma = 1.0)
             double sum_tollerance = 1.0e-9;
-            double include_tollerance = 1.0e-6;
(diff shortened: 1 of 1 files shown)


fevangelista:
@jeffschriber: can you please test this branch to see if the problem with AVAS is solved?

jeffschriber:
I cannot get this test to fail after ~20 or so runs, seems to be fixed!

fevangelista:
Ok, I rebased the PR. Please review once all tests pass.

PR #153, opened 2019-06-26
Test more parts of the code

fevangelista:
## Description
This PR tests more parts of the DSRG-MRPT3 code.

## User Notes
- [x] Adds more test for DSRG-MRPT3

## Checklist
- [x] Ready to go!

### DIFF
--- a/tests/methods/dsrg-mrpt3-9/input.dat
+++ b/tests/methods/dsrg-mrpt3-9/input.dat
@@ -35,6 +35,8 @@ set forte{
    correlation_solver     dsrg-mrpt3
    dsrg_mrpt3_batched     true
    ignore_memory_warnings true
+   int_type CHOLESKY
+   CHOLESKY_TOLERANCE     1e-12
    frozen_docc            [1,0,0,0]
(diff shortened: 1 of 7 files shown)


fevangelista:
@lcyyork: I cut off some code that was not called and tried to test more stuff here. Do you have an example that triggers the use of the batched algorithm using just a bit more than 256 MB of memory and runs quickly? That way we can avoid the extra options that I added to test the code.

lcyyork:
@fevangelista The test case of dsrg-mrpt3-3 is supposed to test the batching and stuff. I think a DF computation of H6 or H8 using vQZ or vTZ basis might be able to do what you asked for.

tests/methods/dsrg-mrpt3-9/input.dat
```diff
@@ -0,0 +1,56 @@
+#! Generated using commit GITCOMMIT
```
lcyyork: Just FYI, this test case does the same test (I assume) as dsrg-mrpt3-3.

PR #158, opened 2019-07-31
Templated Determinant class

fevangelista:
## Description
This PR introduces a new templated Determinant class with variable size (specified at compile time). It removes the `STLBitsetDeterminant` and `UI64Determinant` classes, and includes some of the tricks used in `UI64Determinant` to optimize operations. All the functionality is now handled by the `BitArray<size_t N>` template class.  Determinant functionality is implemented by the `DeterminantImpl<size_t N>` class and in the code we use the alias `Determinant = DeterminantImpl <size_t N>` with a fixed value of `N`.

This PR also adds a residual convergence check to the Davidson-Liu solver and modifies the default number of vectors to be 5 (this improves convergence of FCI,e.g.). Previously we used only the energy. After these changes several tests broke because the CI wave functions were not converged. Updated energy values in tests were obtained from the `master` branch by pushing energy convergence to 10^{-12} - 10^{-14}.

## User Notes
- [x] Add `BitArray<size_t N>` template class
- [x] Added the `DeterminantImpl <size_t N>` template class (to handle a/b string functionality)
- [x] Implemented residual convergence criterion in Davidson-Liu
- [x] Requires support for `if constexpr` (C++17)
- [x] Update ``output.ref` for changed test cases (not done yet)

## Checklist
- [x] Added tests of diagonalization algorithmsm
- [x] Documented source code
- [x] Ready to go!


src/sparse_ci/bitarray.hpp
```diff
@@ -0,0 +1,366 @@
...
+            }
+        }
+        return ~uint64_t(0);
+    }
+
+    /// Clear the first bit set to one
+    void clear_first_one() {
+        for (size_t n = 0; n < nwords_; n++) {
+            // find the first word != 0
+            if (words_[n] != uint64_t(0)) {
```
tyzhang1993: Better use `word_t(0)` consistently instead of `uint64_t(0)`

src/sparse_ci/determinant.hpp
```diff
@@ -0,0 +1,650 @@
...
+            // specialization for 64 + 64 bits
+            return ui64_sign(words_[1], n, m);
+        } else {
+            return slater_sign(n + beta_bit_offset, m + beta_bit_offset);
+        }
+    }
+
+    /// Return the sign for a quadruplet of alpha second quantized operators
+    /// a+(a) a+(b) a(j) a(i) applied to this determinant
+    /// Untested, needs documentation
```
tyzhang1993: I remember I did a lot of work to derive all of these signs for double excitations. I tested and these should be reliable.

fevangelista: Yes, these are just the functions that you wrote copied in the code. I think we already used them in the other determinant classes, so I don't see any potential problem here.

src/sparse_ci/determinant_functions.hpp
```diff
@@ -0,0 +1,154 @@
...
+                                          const std::shared_ptr<ActiveSpaceIntegrals>& ints);
+
+Determinant common_occupation(const Determinant& lhs, const Determinant& rhs);
+
+Determinant different_occupation(const Determinant& lhs, const Determinant& rhs);
+
+/// Find the spin orbitals that are occupied only one determinant (performs a bitwise OR, |)
+Determinant union_occupation(const Determinant& lhs, const Determinant& rhs);
+
+void enforce_spin_completeness(std::vector<Determinant>& det_space, int nmo);
```
jeffschriber: Where does this function go?

PR #161, opened 2019-10-09
Do not store an IntegralTransform object in the ForteIntegrals class

fevangelista:
## Description
This PR fixes a problem where a psi4 DPD buffer was never properly closed

## Checklist
- [x] Documented source code
- [x] Ready to go!

### DIFF
--- a/src/integrals/conventional_integrals.cc
+++ b/src/integrals/conventional_integrals.cc
@@ -41,7 +41,7 @@
 
 #include "conventional_integrals.h"
 
-#define ID(x) integral_transform_->DPD_ID(x)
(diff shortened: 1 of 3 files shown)


fevangelista:
@lcyyork: this should be a quick review.

fevangelista:
This avoids a problem with jupyter notebooks. The previous code did not close a DPD buffer and I think it was because we build and IntegralTransform object and never destroy it. It turns out that we always create a new object when we perform a transformation, so there is no need to keep a smart pointer to IntegralTransform. There is still a problem with running multiple forte jobs, the PSIO file in libmints, but that is a psi4 problem.

PR #162, opened 2019-11-25
CASSCF gradient in forte

ustcwsh:
## Description
implementation of CASSCF in forte

## Notes
- [x] Features added

## Checklist
- [x] Added casscf_gradient.cc under casscf/
- [x] Added additional essential variables and functions in casscf.h under casscf/
- [x] Add option 'DERTYPE' in forte
- [x] Add gradient implementation in forte_old_methods.cc 
- [x] Add the function 'gradient_forte' in pymodule.py
- [x] Create the directory 'gradient_tpdm' under src/ 
- [x] Ready to go!


src/casscf/casscf_gradient.cc
```diff
@@ -0,0 +1,477 @@
...
+namespace forte 
+{
+
+
+/**
+ * Initialize global variables. 
+ */
+void CASSCF::set_all_variables()
+{
+    // Set MOs containers.
```
lcyyork: I wonder if we could put this part outside CASSCF and MRDSRG, since this part seems common for both codes. The thing I am not sure is that the CASSCF may be completely revamped in the future and thus may not deserve much changes at this point.

src/casscf/casscf_gradient.cc
```diff
@@ -0,0 +1,477 @@
...
+    BTF_= std::make_shared<BlockedTensorFactory>();
+    std::string acore_label_ = "c";
+    std::string aactv_label_ = "a";
+    std::string avirt_label_ = "v";
+    std::string bcore_label_ = "C";
+    std::string bactv_label_ = "A";
+    std::string bvirt_label_ = "V";
+
+    // Add Ambit index labels.
+    BTF_->add_mo_space(acore_label_, "mn$%", core_mos_, AlphaSpin);
```
lcyyork: Can you change core labels to m, n, c0, c1, c2, ... instead of characters like !@# and make changes to other labels as well. I know you did not start this but I do think they look hideous and hard to interpret.

src/casscf/casscf_gradient.cc
```diff
@@ -0,0 +1,477 @@
...
+#include "sci/aci.h"
+#include "fci/fci_solver.h"
+#include "base_classes/active_space_solver.h"
+
+#include "sci/fci_mo.h"
+#include "orbital-helpers/mp2_nos.h"
+#include "orbital-helpers/orbitaloptimizer.h"
+#include "orbital-helpers/semi_canonicalize.h"
+
+#ifdef HAVE_CHEMPS2
```
lcyyork: Please clean up the headers and clang-format the file.

src/casscf/casscf_gradient.cc
```diff
@@ -0,0 +1,477 @@
...
+
+    outfile->Printf("Done");
+}
+
+
+/**
+ * Write spin_dependent two-RDMs coefficients using IWL.
+ *
+ * Coefficients in d2aa and d2bb need be multiplied with additional 1/2! 
+ */
```
lcyyork: Thanks for the implementation note. I do prefer a more detailed version of this about all spin cases and maybe give an example. I think there is a factor of something for d2ab as well.

lcyyork:
One more thing, please investigate why TravisCI fails.

lcyyork:
And please at least one test case.

pymodule.py
```diff
@@ -398,6 +398,97 @@ def run_forte(name, **kwargs):
     psi4.core.set_scalar_variable('CURRENT ENERGY', energy)
     return ref_wfn
 
+
+
+def gradient_forte(name, **kwargs):
```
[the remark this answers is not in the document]
ustcwsh: I am not sure since some part has been changed and some functions in run_forte are unnecessary for the gradient computation.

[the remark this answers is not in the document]
lcyyork: I think it is as simple as:

set dertype
call run_forte
contract with derivative integrals

src/casscf/casscf.cc
```diff
@@ -65,6 +65,8 @@ CASSCF::CASSCF(StateInfo state, size_t nroot, std::shared_ptr<SCFInfo> scf_info,
     startup();
 }
 
+
```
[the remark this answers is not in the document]
ustcwsh: This is kelvin's code and there are many commented functions in it. Do we need to remove those functions?

src/casscf/casscf_gradient.cc
ustcwsh: I have fixed this.

src/casscf/casscf_gradient.cc
ustcwsh: The header has been cleaned.

src/casscf/casscf_gradient.cc
ustcwsh: I have added more detailed comments to it.

src/casscf/casscf_gradient.cc
ustcwsh: Maybe extract the procedure of orbital optimizing out of both codes?

fevangelista:
@ustcwsh: I pushed some of the changes to the master branch so that this can be merged. However, can you revert the gradient code to use the reference trick to force psi4 to do the correct computation? We need to use this solution temporary because we cannot update yet the psi4 conda on Travis.

ustcwsh:
> @ustcwsh: I pushed some of the changes to the master branch so that this can be merged. However, can you revert the gradient code to use the reference trick to force psi4 to do the correct computation? We need to use this solution temporary because we cannot update yet the psi4 conda on Travis.

ustcwsh:
I have fixed it.

ustcwsh:
@fevangelista I have reverted to the previous commit where your trick is not existing.

PR #166, opened 2019-12-10
remove timer casscf grad

lcyyork:
## Description
Remove the casscf gradient test case timer.

## Checklist
- [x] Ready to go!


ustcwsh:
marvelous

fevangelista:
Looks good. Next we should cleanup the output. I see some timing printed in `output.dat` and some irregular formatting towards the end. Just minor stuff.

PR #167, opened 2019-12-17
ForteCon Tutorials and interface

fevangelista:
## Description

## User Notes
- [ ] Features added

## Checklist
- [ ] Documented source code
- [ ] Ready to go!


lcyyork:
I think it would be better to use something similar to psi4's `ao_eri` function from the python side that automatically reshape to the correct dimension. I think the current solution of returning pair of data and dimension is not ideal, especially for ForteIntegrals, because the user knows the dimension when he/she passes in the index ranges and thus know the dimension.

fevangelista:
I have found a way to convert ambit::Tensor -> numpy and used this to return ints/RDMs.

fevangelista:
@lcyyork: Can you move `StateInfo::repr()` to the py side? And if you have a chance can you follow the example of `/src/api/api_rdms.cc` and create a separate file to bind `StateInfo`? I am hoping doing this will reduce compile time.

lcyyork:
@fevangelista Sure thing. I will do it after the forum, which takes place this weekend. For the compilation time problem I mentioned several months ago, I suspect that I was using two types of omp (maybe 1 from mkl, 1 from gcc or brew) and compiling ambit code was the problem. Now I specify omp library to be mkl and compiling forte only takes 1 minute on my computer.

fevangelista:
@lcyyork: I am ready to merge this. Can you approve please? You will need to update the tutorial you wrote to use the new tensor interface.

src/integrals/integrals.cc
```diff
@@ -200,6 +200,22 @@ double ForteIntegrals::oei_b(size_t p, size_t q) const {
     return one_electron_integrals_b_[p * aptei_idx_ + q];
 }
 
+ambit::Tensor ForteIntegrals::oei_a_block(const std::vector<size_t>& p,
```
lcyyork: Are these two function necessary? We can probably directly return a copy of py::array.

fevangelista: Well, this way we get both the C++ functionality (via ambit) and the python one (via py::array).

fevangelista:
When this passes we can merge it. For @lcyyork  and @ustcwsh: to make the tests pass I had to revert to the old way we compute gradients where we have to trick the driver. Once we fix the issue with the new conda package and pybind11 we can use the better solution.

PR #175, opened 2020-02-13
Test case fixes for DSRG DIIS

lcyyork:
## Description
Fix several long test cases regarding to the PR of DSRG DIIS.

## User Notes
- [x] Lower the threshold for `mrdsrg-ldsrg2-df-seq-3`
- [x] Add reference output and comparisons for `mrdsrg-ldsrg2-df-4`
- [x] Fix input keyword for `mrdsrg-srgpt2-2`
- [x] Adapt the `active_space_solver` and `correlation_solver` for the unused MRDSRG test cases.

## Checklist
- [x] Ready to go!


lcyyork:
@fevangelista I have run all the mrdsrg-ldsrg2 and mrdsrg-pt2 test cases. Most of them are in the unused group. However, I found `mrdsrg-ldsrg2-df-4` is never compared against the reference values. Moreover I have lowered the comparison precision to 1.0e-8 because I found the nuclear repulsion energies differ comparing to the reference output values.

PR #182, opened 2020-02-25
Spin adapted dsrg

lcyyork:
## Description
Spin-adapted DSRG implementation.

## User Notes
- [x] Fix a bug of state averaging in `ActiveSpaceSolver`.
- [x] Add supports of spin-averaged RDMs in `RDMs` and `ActiveSpaceSolver` classes.
- [x] Implement the spin-adapted single commutators up to singles and doubles.
- [x] Add complete supports of PT2, PT3, LDSRG(2) energies.
- [x] Support non-semicanonical orbitals and completely zeroing CCVV block.

The spin-adapted DSRG code is arranged in the following way:
```
SADSRG --|
         -- SA_MRDSRG (nonperturbative DSRG implementation)
         |
         -- SA_DSRGPT --|
                        -- SA_MRPT2 (DSRG-MRPT2 implementation)
                        |
                        -- SA_MRPT3 (DSRG-MRPT3 implementation)
```

## Checklist
- [x] Add several tests under groups `mrdsrg-spin-adapted`, `mrdsrg-spin-adapted-pt2`, and `mrdsrg-spin-adapted-pt3`.
- [x] Ready to go!

## TODOs
- [ ] Implement internal amplitudes for general PT
- [ ] Test speed for PT2 threading
- [ ] The test cases' numbers may be updated when Psi4's CASSCF problem is fixed.


lcyyork:
@henankf223 @fevangelista This PR is not ready because the actual spin-adapted DSRG is not yet implemented. However, I made changes to ActiveSpaceSolver and RDMs to average spin multiplets. Now, the spin-integrated DSRG code should be able to handle any spin cases by setting `SPIN_AVG_DENSITY` to `True`.

fevangelista:
@lcyyork: thanks! I will download a copy and start to go over it!

docs/sphinx/methods/dsrg.rst
```diff
@@ -16,10 +16,18 @@ Driven Similarity Renormalization Group
...
-  For those in desperate need to perform computations on doublets, an alternative way is to add a hydrogen atom far away from the system and perform a singlet computation.
-  Spin adaptation is on the TODO list.
+  The examples used in this manual are written based on the spin-integrated code
+  that works only for singlet molecules and those with odd multiplicities.
+  To invoke the spin-adated implementation (that works for general multiplicities),
+  the user needs to specify the following keywords:
+  ::
+
+     correlation_solver  sa-mrdsrg  # spin-adapted DSRG computation
+     spin_avg_density    true       # must use spin-averaged densities
```
fevangelista: Could we get rid of the variable `spin_avg_density`? At least we should have logic to enforce that when `sa-mrdsrg` is set then `spin_avg_density = True`.

pymodule.py
```diff
@@ -270,7 +274,11 @@ def compute_dsrg_unrelaxed_energy(correlation_solver_type, rdms, scf_info, optio
                                   ints, mo_space_info, Ua, Ub):
     Heff_actv_implemented = False
 
-    if correlation_solver_type == "MRDSRG_SO":
+    if correlation_solver_type == "SA-MRDSRG":
```
fevangelista: Here is a good spot for testing that `spin_avg_density` is set to true.

src/base_classes/active_space_solver.cc
```diff
@@ -70,42 +73,58 @@ const std::map<StateInfo, std::vector<double>>& ActiveSpaceSolver::compute_energ
             method_, state, nroot, scf_info_, mo_space_info_, as_ints_, options_);
         state_method_map_[state] = method;
 
+        int twice_ms = state.twice_ms();
+        if (twice_ms < 0 and ms_avg_) {
+            psi::outfile->Printf(
+                "\n  Continue to next symmetry block: No need to compute for ms = %d / 2 < 0.",
```
fevangelista:
```suggestion
                "\n  Continue to next symmetry block: No need to find solution for ms = %d / 2 < 0.",
```

src/base_classes/active_space_solver.cc
```diff
@@ -70,42 +73,58 @@ const std::map<StateInfo, std::vector<double>>& ActiveSpaceSolver::compute_energ
...
     }
     print_energies(state_energies_map_);
     return state_energies_map_;
 }
 
 void ActiveSpaceSolver::print_energies(std::map<StateInfo, std::vector<double>>& energies) {
     print_h2("Energy Summary");
-    psi::outfile->Printf("\n    Multi.  Irrep.  No.               Energy");
-    std::string dash(41, '-');
+    psi::outfile->Printf("\n    Multi.(ms)  Irrep.  No.               Energy");
```
fevangelista: Is this M_S? If yes we can just replace 
```suggestion
    psi::outfile->Printf("\n   M_S    Irrep.  No.               Energy");
```

src/base_classes/active_space_solver.cc
```diff
@@ -142,23 +161,23 @@ void ActiveSpaceSolver::print_options() {
         nstates += state_nroot.second;
     }
 
-    int ltotal = 6 + 2 + 6 + 2 + 7 + 2;
+    int ltotal = 6 + 2 + 10 + 2 + 6;
     std::string dash(ltotal, '-');
-    psi::outfile->Printf("\n    Irrep.  Multi.  Nstates");
+    psi::outfile->Printf("\n    Irrep.  Multi.(ms)      N");
```
fevangelista: M_S?

src/base_classes/active_space_solver.cc
```diff
@@ -184,16 +203,33 @@ make_state_weights_map(std::shared_ptr<ForteOptions> options,
...
 
     if (avg_state.size() == 0) {
-
         int nroot = options->get_int("NROOT");
         int root = options->get_int("ROOT");
 
-        std::vector<double> weights(nroot, 0.0);
-        weights[root] = 1.0;
-        state_weights_map[state] = weights;
+        if (!options->get_bool("SPIN_AVG_DENSITY")) {
```
fevangelista:
```suggestion
        if (not options->get_bool("SPIN_AVG_DENSITY")) {
```

src/base_classes/active_space_solver.cc
```diff
@@ -184,16 +203,33 @@ make_state_weights_map(std::shared_ptr<ForteOptions> options,
...
     py::list avg_state = options->get_gen_list("AVG_STATE");
 
     if (avg_state.size() == 0) {
-
         int nroot = options->get_int("NROOT");
         int root = options->get_int("ROOT");
 
-        std::vector<double> weights(nroot, 0.0);
-        weights[root] = 1.0;
-        state_weights_map[state] = weights;
```
fevangelista: Can you add a comments to the code to explain what is going on?

src/base_classes/active_space_solver.cc
```diff
@@ -284,38 +338,40 @@ make_state_weights_map(std::shared_ptr<ForteOptions> options,
 
 RDMs ActiveSpaceSolver::compute_average_rdms(
```
fevangelista: I think I would like for this function to be broken into three parts, with `compute_average_rdms()` calling two specialized functions for the averaged and non-averaged cases.

src/base_classes/active_space_solver.h
```diff
@@ -135,6 +135,9 @@ class ActiveSpaceSolver {
     /// A map of state symmetries to vectors of computed energies under given state symmetry
     std::map<StateInfo, std::vector<double>> state_energies_map_;
 
+    /// Average spin multiplets for RDMs
```
fevangelista: Can you say what happens in the code if this is set to true or false?

src/base_classes/rdms.cc
```diff
@@ -86,30 +146,92 @@ ambit::Tensor RDMs::L3aaa() {
...
-        L3bbb_ = g3bbb_.clone();
-        make_cumulant_L3bbb_in_place(g1b_, L2bb_, L3bbb_);
+        L3bbb_ = g3bbb().clone();
+        make_cumulant_L3bbb_in_place(g1b(), L2bb(), L3bbb_);
         have_L3bbb_ = true;
     }
     return L3bbb_;
 }
 
+ambit::Tensor RDMs::SFg2() {
```
fevangelista: Can you give the precise definition of what this returns in the header file?

src/base_classes/rdms.h
```diff
@@ -81,33 +81,40 @@ class RDMs {
...
     /// @return the alpha-beta-beta 3-RDM
-    ambit::Tensor g3abb() const { return g3abb_; }
+    ambit::Tensor g3abb();
     /// @return the beta-beta-beta 3-RDM
-    ambit::Tensor g3bbb() const { return g3bbb_; }
+    ambit::Tensor g3bbb();
 
     // Spin-free RDMs
 
     /// @return the spin-free 2-RDM
```
fevangelista: Here add definition of what does this return.

src/base_classes/rdms.h
```diff
@@ -191,8 +232,33 @@ class RDMs {
...
+
+    /// Was SF_L1_ built?
+    bool have_SF_L1_ = false;
+    /// Was SF_L2_ built?
+    bool have_SF_L2_ = false;
+    /// Was SF_L3_ built?
+    bool have_SF_L3_ = false;
 };
 
+/**
```
fevangelista: This is great!

src/mrdsrg-helper/dsrg_mem.h
```diff
@@ -0,0 +1,97 @@
...
+
+#include <vector>
+#include <string>
+#include <map>
+
+#include "base_classes/mo_space_info.h"
+
+namespace forte {
+
+class DSRG_MEM {
```
fevangelista: Can this be used for other codes? If it can then let's not be specific (DSRG). It might be nice to start it outside of the DSRG codes. However, this might not play well with python.

src/mrdsrg-helper/dsrg_mem.h
```diff
@@ -0,0 +1,97 @@
...
+#define _dsrg_mem_h_
+
+#include <vector>
+#include <string>
+#include <map>
+
+#include "base_classes/mo_space_info.h"
+
+namespace forte {
+
```
fevangelista: What does this class do?

src/base_classes/active_space_solver.cc
lcyyork: No, there are two values here. One is for multiplicity (i.e., Multi.) and the other is for MS that is in the parentheses.

src/base_classes/active_space_solver.cc
lcyyork: Same as the last one.

docs/sphinx/methods/dsrg.rst
lcyyork: OK, let me work on it.

src/mrdsrg-helper/dsrg_mem.h
lcyyork: This class is just for memory inspection before any actual DSRG computation. So this class might be useful for other dynamical correlation solver.

src/mrdsrg-helper/dsrg_mem.h
lcyyork: This class is created to check/print the memory usage before any real DSRG computation. Different DSRG levels have different memory requirement and it is useful to terminate the job if it lacks of memory.

Basically the user sets the max memory available and a map of the space character to its size (e.g., 'c': 20, 'a': 4, 'v': 80, 'h': 24, 'p': 84, 'g': 104).

Take LDSRG(2) as an example. It stores F, V, T1, T2, Hbar1, Hbar2, and other intermediates. To compute the memory usage of T1 and T2, the user can `add_entry` and pass the name and the labels: `add_entry("T1 and T2 amplitudes", {"hp", "hhpp"});`.

There are two sets of global intermediates used in LDSRG(2) both of sizes g^2 + g^4 and this can be added as `add_entry("Global intermediates", {"gg", "gggg"}, 2);`.

Besides global memory usage, there are also local usage. In this case, the max local memory usage will be the bottleneck. To add a local memory usage, use `add_entry("Local storage for S1 and S2 amplitudes", {"hp", "hhpp"}, 1, false);`

The `print` function will print all passed in entries and terminate the job if the global memory plus the max local memory is greater than the max available set by the user.

docs/sphinx/methods/dsrg.rst
```diff
@@ -16,10 +16,27 @@ Driven Similarity Renormalization Group
...
-  The current implementation does not employ spin-adapted equations and it does not work for even spin multiplicities (doublet, quartets, etc.).
-  For odd multiplicities, we assume low-spin configurations (by default, no need to specify in the input file).
-  For those in desperate need to perform computations on doublets, an alternative way is to add a hydrogen atom far away from the system and perform a singlet computation.
-  Spin adaptation is on the TODO list.
+  The examples used in this manual are written based on the spin-integrated code.
+  To make the spin-integrated code work properly for molecules with even multiplicities,
+  the user should specify the following keyword:
+  ::
+
+     spin_avg_density    true       # use spin-summed reduced density matrices
```
fevangelista: Is this obsolete now? If it is, it might still be a good idea to mention it somewhere, either here or in a section on active space solvers.

fevangelista: BTW, how do you feel about moving our manual to jupyter?

fevangelista:
@lcyyork: can you run the full test suite and make sure everything works?

docs/sphinx/methods/dsrg.rst
lcyyork: No, it is not obsolete. It shows the user should specify this keyword to run jobs with even multiplicities using the spin integrated code. I will emphasize this using bold font.

I am OK to move them to Jupyter but not sure what the benefit is. I assume they will still be markdown in Jupyter.

lcyyork:
@fevangelista All tests pass on my computer.

PR #188, opened 2020-05-13
Add disk I/O support for BlockedTensor

lcyyork:
## Description
- Add disk I/O support for BlockedTensor.
- Add restart functionality for spin-adapted/integrated MR-LDSRG(2) relaxation.
- Add read/write amplitudes for spin-adapted/integrated MR-DSRG.
- Completely rewrite the DSRG solver in Python side.
- Rewrite part of `aci_build_F.cc` and `gasci_build_F.cc` to make them run with OpenMP.
- TravisCI now runs on MacOS and only run test cases labeled as short.

## User Notes
- [x] Add restart functionality for spin-adapted/integrated MR-LDSRG(2) relaxation.
- [x] Add read/write amplitudes for spin-adapted/integrated MR-DSRG.

## Checklist
- [x] Ready to go!


lcyyork:
The functionality of writing a BlockedTensor to disk may interest you. @henankf223 @fevangelista

fevangelista:
@lcyyork : this looks like some great new functionality. Can any of this be put in ambit so that it more widely available? @tyzhang1993 may also have some suggestions.

lcyyork:
> @lcyyork : this looks like some great new functionality. Can any of this be put in ambit so that it more widely available? @tyzhang1993 may also have some suggestions.

Yes, I think the disk read/write can be adapted to ambit for core Tensor/BlockedTensor.

tyzhang1993:
This is definitely a good feature. Putting the functionality in Ambit is more reasonable from the design point of view because Ambit actually holds the data.
I am wondering if those tensor blocks can be gathered to a single bin file rather than one for each block?

lcyyork:
The reason that a BT is separated into blocks here is because I reuse the old code of writing a vector of double [basically a Tensor.data()] in forte. The size check can be easily done for each block since the writing function first dumps the number of elements and then the actual data. This also gives the flexibility that one can read different blocks from different calculations.

However, I would have to agree generating a large number of temporary files does not look very good. Putting everything in a single binary seems good choice.

lcyyork:
> I saw that the test cases related to travis-CI fail.

It is because the test cases run out of Travis-CI time limit. I removed a few pt2 tests cases to long. Hope this works.

lcyyork:
@fevangelista @tyzhang1993 Any suggestions on how to condense the number of files?

lcyyork:
@fevangelista I have adapted the save/load functions from Ambit but it seems the it is not ready from the python side (I do not see ambit install in the ymal file but I think ambit is installed when doing conda psi4).

fevangelista:
@lcyyork : I think we had a misunderstanding. My suggestion was to leave all the functions to read/write tensors, but to conform them to the standard used in ambit. So that when a conda package is ready we can switch all calls to `forte::load/save` to `ambit::load/save`.

lcyyork:
I think I did not. The forte::save/load has to have exactly the same function arguments to the ambit::save/load. The forte::save/load have not been implemented now.

lcyyork:
@loriab Hi Lori, do you have any idea the `TypeError` caused by pybind11 using Psi4 from conda (e.g. `0177a79`)? If I compile Psi4 myself, this error will go away (e.g., `5de1c1b`).

lcyyork:
@fevangelista Please take a look of this PR. The codecov will be slightly reduced because the test cases now shrink to 71.

src/helpers/disk_io.cc
```diff
@@ -0,0 +1,194 @@
+/*
```
fevangelista, line 1: @lcyyork : are these still needed? I assume that the I/O in ambit should be enough now.

docs/sphinx/methods/dsrg.rst
```diff
@@ -679,7 +679,43 @@ We have implemented the following choices for the zeroth-order Hamiltonian.
...
+and ``forte.mrdsrg.adapted.t1.bin`` for spin-adapted code (i.e., `correlation_solver` set to `sa-mrdsrg`).
+To read amplitudes in the current directory (must follow the same file name convention),
+the user needs to invoke the ``DSRG_READ_AMPS`` keyword.
+
+.. note::
+  In general, we should make sure the orbital phases are consistent between reading and writing amplitudes.
+  For example, the following shows part of the input to ensure the coefficient of the first AO being positive for all MOs. ::
+
+    ...
+    Escf, wfn = energy('scf', return_wfn=True)
```
fevangelista, line 699: This procedure should be implemented in forte. (we might have to coordinate with Nan's changes to the QC repo).

src/integrals/integrals.cc
```diff
@@ -245,6 +245,84 @@ void ForteIntegrals::set_oei(size_t p, size_t q, double value, bool alpha) {
     p_oei[p * aptei_idx_ + q] = value;
 }
 
+void ForteIntegrals::fix_orbital_phases(std::shared_ptr<psi::Matrix> U, bool is_alpha, bool debug) {
```
fevangelista, line 248: Where is this called?

docs/sphinx/methods/dsrg.rst
fevangelista: Ok, I found some C++ code below. So maybe things just need to be harmonized.

src/helpers/disk_io.cc
lcyyork: > @lcyyork : are these still needed? I assume that the I/O in ambit should be enough now.

I moved the read/write a vector of double to this file. These two functions were used to read and write RDMs before. I guess we no longer need these two functions.

src/integrals/integrals.cc
lcyyork: in Semicanonical class

PR #189, opened 2020-05-14
Fix bug in iterative solver

fevangelista:
## Description
This PR fixes a bug in the Davidson-Liu solver. The added test case make sure that the solver correctly converges to a triplet state found above a singlet. As usual, this started like a PR that was supposed to solve one problem and it ended up addressing a bunch of little extra things.

## Features
- [X] Fix issue with parsing of relativistic basis
- [X] Move some common variables from ACI to SCI class
- [X] Read one-electron integrals from wavefunction
- [X] Improvements to the DL solver
- [X] Expose some options

## Checklist
- [X] Added tests of new features
- [X] Removed comments in code and input files
- [X] Documented source code
- [x] Ready to go!

### DIFF
--- a/register_forte_options.py
+++ b/register_forte_options.py
@@ -443,7 +443,7 @@ def register_aci_options(options):
 
     options.add_bool("ACI_PRINT_REFS", False, "Print the P space?")
 
-    options.add_int("DL_GUESS_SIZE",26,
+    options.add_int("DL_GUESS_SIZE",50,
                           "Set the initial guess space size for DL solver")
 
     options.add_int("N_GUESS_VEC", 10,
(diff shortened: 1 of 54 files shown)


register_forte_options.py
```diff
@@ -421,13 +424,13 @@ def register_aci_options(options):
 
     options.add_int("SCI_MAX_CYCLE", 20, "Maximum number of cycles")
 
-    options.add_bool("ACI_QUIET_MODE", False,
+    options.add_bool("SCI_QUIET_MODE", False,
```
lcyyork, line 427: Should these moved to the sci options block?

fevangelista: Thanks for catching this.

PR #191, opened 2020-06-22
fix cas when there is no restricted_docc

lcyyork:
## Description
Same as title.

## User Notes
- [x] Changes `CASSCF_ITERATIONS` to `CASSCF_MAXITER` to standardize with other parts of Forte

## Implementation Notes
- [x] No longer transform all integrals for each iteration

## Checklist
- [x] Added test case `casscf-9` where there is no `frozen_docc` nor `restricted_docc`
- [x] Documented source code
- [x] Ready to go!


lcyyork:
@fevangelista @huangm29 This should fix the CASSCF energy blowing up problem when there is no restricted_docc.

fevangelista:
@lcyyork : I'll test it right away.

fevangelista:
I pushed some changes that I wanted to do. I also found code that was redundant. Right now I would like to try to re-enable the no-printing mode, so that you only see the CASSCF iterations.

fevangelista:
Maybe we should go one step further and move CASSCF out of forte_old_methods and send it to the python side. I am not sure how to trigged CASSCF, maybe we just need a option like `optimize_mos` or `do_casscf`.

lcyyork:
Let me merge some of my changes in the casscf_clean_up branch to this branch.

fevangelista:
I think I am done with my changes. I could not yet fix the printing issue but that's a minor esthetic issue.

lcyyork:
So I merge my casscf_clean_up branch changes to here. Basically I avoid the full integral transformation for each macro iteration since only a tiny fraction of integrals are needed.

lcyyork:
I am finally done with the changes. Let us wait and see what TravisCI says.

lcyyork:
@henankf223 I add Fock build in CASSCF now. You may want to take a look if you have not yet make it work.

PR #192, opened 2020-07-03
Read integrals from FCIDUMP

### LINKED ISSUE #181: Improvements to the Integral classes
Goals for a new PR

- [ ] Develop the `CustomIntegrals` objects and introduce the ability to read integrals stored on disk I various formats (including FCIDUMP).

- [ ] Introduce separate classes for integrals based on 3- or 4-dimensional tensors. Currently we derive all classes from `ForteIntegrals` as follows:
```
ForteIntegrals --|
                 -- ConventionalIntegrals
                 |
                 -- DFIntegrals
                 |
                 -- DiskDFIntegrals
                 .
                 .
                 .
```
This requires the `ForteIntegrals` interface to be too general. A better way would be to separate integrals based on 3- and 4- indices and write a consistent interface in two separate classes `FourIndexIntegrals` and `ThreeIndexIntegrals`
```
ForteIntegrals --|
                 -- FourIndexIntegrals  --|
                 |                        -- ConventionalIntegrals
                 |                        |
                 |                        -- CustomIntegrals
                 |                        
                 -- ThreeIndexIntegrals --|
                                          -- DFIntegrals
                                          |
                                          -- DiskDFIntegrals
                                          .
                                          .
                                          .
```

fevangelista:
## Description
Let Forte start a computation from a FCIDUMP file. Closes #181.

This PR lets users read integrals from a file in the FCIDUMP format. For example, if the integrals are written to disk (`INTDUMP` file) then a forte input file can be condensed to
```
import forte

set forte {
  int_type            fcidump
  active_space_solver fci
  frozen_docc         [2 ,0 ,0 ,0]
  restricted_docc     [2 ,0 ,0 ,0]
  active              [2 ,2 ,2 ,2]
  e_convergence       12
}

energy('forte')
```

## User Notes
- [x] Implements reading of FCIDUMP files
- [x] Implements `CustomIntegrals` class (some functionality is not complete)
- [x] Some of the python interface files were changed to avoid using psi4. For example, the interface to make a MOSpaceInfo object in python now requires the user to provide a Dimension object instead of a psi4 `Wavefunction`
```
    nmopi = wfn.nmopi()
    mo_space_info = forte.make_mo_space_info(nmopi, forte_options)
```
- [x] Removes some unnecessary elements of `SCFInfo` (the goal is eliminate this class at some point)
- [x] Introduces `None` default values in the `ForteOption` class for quantities that do not have a default (e.f. charge/multiplicity)

## Checklist
- [x] Documented source code
- [ ] Ready to go!


proc/fcidump.py
```diff
@@ -0,0 +1,460 @@
...
+    Additional one-electron integrals, including orbital energies, can also be saved.
+    This latter format can be used with the HANDE QMC code but is not standard.
+    This function converts the irrep labels from Cotton's order (used in psi4)
+    to molpro ordering. Since the latter is ambiguous unless the point group is
+    specificied, this function has an option to write the point group label in the
+    FCIDUMP file. This, however, is not part of the standard.
+
+    :returns: None
+
+    :raises: ValidationError when SCF wavefunction is not RHF
```
lcyyork, line 52: Is this still valid?

proc/fcidump.py
```diff
@@ -0,0 +1,460 @@
...
+    eri = np.zeros((nbf, nbf, nbf, nbf))
+    eri[(idxs[sl, 0], idxs[sl, 1], idxs[sl, 2], idxs[sl, 3])] = ints[sl]
+    eri[(idxs[sl, 0], idxs[sl, 1], idxs[sl, 3], idxs[sl, 2])] = ints[sl]
+    eri[(idxs[sl, 1], idxs[sl, 0], idxs[sl, 2], idxs[sl, 3])] = ints[sl]
+    eri[(idxs[sl, 1], idxs[sl, 0], idxs[sl, 3], idxs[sl, 2])] = ints[sl]
+    eri[(idxs[sl, 2], idxs[sl, 3], idxs[sl, 0], idxs[sl, 1])] = ints[sl]
+    eri[(idxs[sl, 3], idxs[sl, 2], idxs[sl, 0], idxs[sl, 1])] = ints[sl]
+    eri[(idxs[sl, 2], idxs[sl, 3], idxs[sl, 1], idxs[sl, 0])] = ints[sl]
+    eri[(idxs[sl, 3], idxs[sl, 2], idxs[sl, 1], idxs[sl, 0])] = ints[sl]
+    intdump['eri'] = eri
```
lcyyork, line 362: I was wondering if something like DASK should be used here since the ERI can be very large.

register_forte_options.py
```diff
@@ -47,17 +49,21 @@ def register_driver_options(options):
                           'The type of computation')
 
     options.add_int(
-        "CHARGE", 0,
+        "NEL", None,
```
lcyyork, line 52: Does it mean every time the user should specify these keywords?

fevangelista: This is a good question and I should put it into the manual. These quantities are relevant to building the StateInfo objects in forte. Charge, multiplicity, etc.  do not have a default because they cannot have one. However, if you run a job via psi4 the code will by default read charge etc. from the wave function object. If charge and other values are specified then they override the values read from psi4::wavefunction. In FCIDUMP there is enough information to reconstruct a StateInfo object so they are also not needed.

proc/fcidump.py
fevangelista: This is just code from psi4. I don't think that FCIDUMP is the format one would use for very large orbital spaces (probably good for < 100 MOs)

proc/fcidump.py
fevangelista: No, this needs to be revised in forte and psi4. Thanks for noticing it.

src/base_classes/scf_info.h
```diff
@@ -32,27 +32,23 @@
...
-// class Vector;
-
 namespace forte {
 
 class SCFInfo {
   public:
     /// Constructor based on Psi4 Wavefunction
     SCFInfo(psi::SharedWavefunction wfn);
 
+    /// Constructor based on Psi4 Wavefunction
```
lcyyork, line 42: Comment seems incorrect.

src/integrals/integrals.h
```diff
@@ -63,7 +63,7 @@ enum class IntegralSpinRestriction { Restricted, Unrestricted };
  *
  * This decides the type of integral used in a Forte computation
  */
-enum IntegralType { Conventional, DF, Cholesky, DiskDF, DistDF, Own, Custom };
+enum IntegralType { Conventional, DF, Cholesky, DiskDF, DistDF, Custom };
```
lcyyork, line 66: Is DistDF functional?

proc/fcidump.py
lcyyork: Thanks for the information. I did not know that.

register_forte_options.py
lcyyork: Thanks, I see those if statements later in the files.

ustcwsh:
I checked the change in CASSCF part and it looks okay to me. BTW, is it essential to change python_ver in env to 3.7? I am currently using 3.6 and not familiar with new functionalities in 3.7, not sure if it affects forte functions

src/integrals/integrals.h
fevangelista: That's a good question. It is possible that we lost all the functionality. I have been wondering if we should remove support for it. However, it would be a pity to go back and forth with MPI.

fevangelista:
@lcyyork : I have an updated version of psi4 and this runs with it (however, some changes must be done in psi4 to enable the output of FCIDUMP files with the point group symmetry information). I am working on a PR now. I have copied some of this code here so we don't need to update psi4 right away. @ustcwsh : yes, it's good to keep in mind the changes that I made to pymodule. Also note that now CASSCF lives on the py-side. No need to use Py 3.7. That's only for TravisCI.

fevangelista:
I'll rerun all the tests and if passes will merge

PR #197, opened 2020-07-30
Add GAS spaces to forte

fevangelista:
## Description
This PR introduces a fundamental change in the way the active orbitals are treated. GAS spaces are introduced and the active orbitals are defined as the composite space of GAS1-6. The user can specify the number of active orbitals in standard CAS computations wither via the GAS1 option or the ACTIVE option. This is possible because now the MOSpaceInfo class can handle assignment to a composite space. For example, if a composite space C = {A, B} is the union of A and B, then if the use sets C = [4,3] then this will result in the assignment A = [4,3] and B = [0,0].
Another important change is that the semicanonicalization procedure is now automated and will diagonalize the Fock matrix in all elementary spaces (excluding frozen orbitals). This feature might also be convenient when running embedding computations.

This PR also allows Forte to run finite difference computations based on CASSCF references from psi4. To specify the orbital type, a new keyword has been added `REF_TYPE`

## User Notes
- [x] Added GAS1-6 spaces
- [x] Modified the MOSpaceInfo class to handle assignment to composite spaces
- [x] Added `REF_TYPE` option to control what type of psi4 wave function should be computed if forte is not given one.
- [x] Added a pytest test case for the MOSpaceInfo class
- [x] Added a test for finite difference computations

## Checklist
- [x] Tested new features
- [x] Documented source code
- [x] Ready to go!


fevangelista:
@lcyyork : this is a good point. However, this was already the case for composite spaces before. I am not what is best. Computationally it might make sense to group orbitals by irrep first. But there might be several parts of forte that make some assumption regarding the ordering of composite spaces.

lcyyork:
@fevangelista I was worried the same thing so I decided to just add an example in the header comments.

fevangelista:
@lcyyork @huangm29 : I changed everything so that now all composite spaces are ordered in Pitzer ordering. That is, orbitals are sorted by irrep, and within an irrep they are sorted by elementary space. All tests in forte (including long ones) pass. I also added function `pos_in_space(space, composite_space)` which can be used to get the position of some orbitals in a given space. I documented MOSpaceInfo extensively. I just need to do another pass on the documentation and then I think this will be ready to be merged.

lcyyork:
@fevangelista Thanks for doing this!

PR #198, opened 2020-08-11
Gas ci

huangm29:
## Description
This PR allows the Generalized Active Space(GAS) method to the CI and SCF calculations. The code will first read how the active space is distributed into different GAS, then reads the maximum and the minimum number of electrons in different GAS and generates either (1) All the possible determinants (2) A single determinant within the GAS. Then the ACI code will iterate within all the determinants that fulfill the GAS electron requirement and calculates the eigenstate. A temporary feature (Keyword: Single_calculation) is added to do the FCI calculation within ACI code rather than setting sigma = 0 and doing the diagonalization twice).

The orbital optimization based on CI/ACI calculation leads to GASSCF calculations. This code contains the Hessian matrix elements between orbitals belong to different GAS, which will result in an orbital optimization between different GAS.

Another feature that this PR adds is to a state-average GAS-ACI (MULTI_GAS) method which allows a state-averaged calculation using different GAS electron settings. This enables a state-average ACI calculation containing, for example, a core-excited state and a ground electronic state. The orbital optimization can be further processed for SA-GASSCF calculation. A keyword: CASSCF_FROZEN_ORBITAL is added to zero-out ALL the possible rotations (setting gradient to zero) that involved a specific orbital in CASSCF, which will make sure the orbital optimization does not collapse the core-excited state to a valence excited state in orbital optimization.

## User Notes
- [ ] GAS-CI (GAS-ACI) feature
- [ ] GASSCF feature
- [ ] MULTI_GAS feature
- [ ] Treating certain orbital as frozen in orbital optimization
- [ ] Added 3 test cases for GAS-CI and 4 test cases for GAS-ACI
- [ ] Added 3 test cases for GASSCF and 3 test cases for SA-GASSCF 
- [ ] Added 1 test case for GASSCF-DSRG-MRPT2 and 1 test case for SA-GASSCF-DSRG-MRPT2

## Checklist
- [ ] Documented source code
- [ ] Ready to go!


fevangelista:
@huangm29 : All of the tests pass. I'd say we go ahead and merge this and then you can work a second PR to clean up some of the GAS routines. Thank you for this contribution to Forte!

PR #200, opened 2020-08-20
New 2-Step MCSCF Code

lcyyork:
## Description
This PR adds a new 2-step MCSCF code in Forte.

- The new 2-step MCSCF consists of macro (solve CI for given orbital coefficients) and micro (optimize orbitals for given CI coefficients) iterations.
- In macro iterations, orbitals are extrapolated using DIIS. In micro iterations, orbital optimization is performed using 
limited-memory BFGS (L-BFGS).
- The number of iterations for L-BFGS minimization is initially set to `CASSCF_MICRO_MINITER` and changes during the macro iterations but it never exceeds `CASSCF_MICRO_MAXITER`.
- Evaluation of orbital gradient and diagonal Hessian is computed in `CASSCF_ORB_GRAD` class.
- The implementation uses `JK` technology of Psi4 (i.e. OK to use PK, DF, DISKDF, CD integrals) and does support `FCIDUMP` format.
- Frozen-core approximation is supported for both energy and gradient.
- Single reference RHF/ROHF can also be computed using the current implementation.
- An orbital orthonormality test is performed when a `ref_wfn` is passed. If the incoming orbitals are not from the current geometry (**C**^T **S** **C** != 1), an SCF calculation and orbital orthonormality procedure will be performed. As such, the incoming orbitals will be treated as starting guess for MCSCF.

## TODO List
- [x] Adapt GAS (semicanonicalization, internal orbital rotation).
- [ ] Add natural orbital support in `SemiCanonical` class.
- [ ] Find better ways to handle mixed orbitals in TPDM transformation (MCSCF gradient with HF frozen-core orbitals).
- [x] Skip ForteIntegral build for CASSCF and store the JK object in ForteIntegral.
- [ ] The CI problem should read the CI coefficients from previous macro iterations. The CI coefficients can also be extrapolated using DIIS.
- [ ] Analytic energy gradient for state-averaged MCSCF. The nuclear gradients target only one of the state.
- [x] Add support of using `FCIDUMP` integrals. This is very inefficient multiple 2-e integral transformation.
- [ ] Optimize for `FCIDUMP` and `Conventional` integrals (only MO two-electron integrals in chemists notation). We can store only the alpha-beta case and the alpha-alpha/beta-beta can be computed when needed.

## User Notes
[...]

PR #205, opened 2020-11-23
Optimizations to spin-adapted DSRG-PT2 using DiskDF integrals

lcyyork:
## Description
The original purpose of this PR is to add more printing for each step of spin-adapted DSRG-PT2.

1. Along the way, a serious limitation of DiskDF integrals  was found: the `three_integral_block` function loops over the second indices of (L|pq) (i.e., p) and stores all p of (L|q) matrices where q always the size of nmo. One can run out of memory if the size of p is large. The major contribution of this PR is to fix this issue.

2. Another important change of thie PR is that `ActiveSpaceIntegrals` now adapts the spin restrictions. For restricted integrals, it only calls alpha-beta spin case of `ForteIntegral` and build alpha-alpha and beta-beta spin cases using the alpha-beta case.

3. New batching functions are implemented for spin-adapted DSRG-PT2 for using DiskDF integrals. The goal is to minimize the number of function calls of `three_integral_block`.

4. A minor change is made to `timer` where it now holds a `local_timer` so that the elapased time is returned when `stop()` is called.

P.S. The history is messed up but this PR follows the commit dea446d of 2-step MCSCF PR #200.

## User Notes
- [x] Add uniform printings functions in sadsrg.
- [x] Add `local_timer` in `timer` and returns the elapsed time when `stop()`.
- [x] Optimize `three_integral_block` for `DiskDFIntegrals`.
- [x] Adapt spin restrictions to `ActiveSpaceIntegrals`.
- [x] Optimize spin-adapted DSRG-PT2 when using `DiskDFIntegrals`.

## Checklist
- [x] Documented source code
- [x] Ready to go!


lcyyork:
All tests of the standard set pass on my computer.

src/mrdsrg-spin-adapted/sa_mrpt2.cc
```diff
@@ -124,24 +125,63 @@ void SA_MRPT2::build_minimal_V() {
...
+        auto size = core_batches[batch].size();
+        auto Bsub = ints_->three_integral_block(aux_mos_, virt_mos_, core_batches[batch]);
 
-        Vsub("uev") = B.block("Laa")("guv") * Bsub("ge");
+        auto Vsub = ambit::Tensor::build(tensor_type_, "Vsub PT2", {na, na, nv, size});
+        Vsub("uvem") = Baa("guv") * Bsub("gem");
 
         Vsub.citerate([&](const std::vector<size_t>& i, const double& value) {
-            Vavac[i[0] * dim1 + i[1] * dim2 + i[2] * nc + c] = value;
+            Vavac[i[0] * dim1 + i[2] * dim2 + i[1] * nc + i[3] + offset] = value;
```
fevangelista, line 146: I noticed a indexing change here. Is this due to a reorganization of the data structure?

src/mrdsrg-spin-adapted/sadsrg.cc
```diff
@@ -260,7 +265,8 @@ void SADSRG::check_init_memory() {
 
     mem_left -= n_ele * sizeof(double);
     if (mem_left < 0) {
-        throw psi::PSIEXCEPTION("Not enough memory to run FORTE.");
+        std::string msg = "Not enough memory to run spin-adapted DSRG. Try DiskDF integrals?";
```
fevangelista, line 268: This is a great function. Can we tell the user how much more memory is needed?

src/mrdsrg-spin-adapted/sadsrg.cc
```diff
@@ -393,11 +364,13 @@ double SADSRG::compute_reference_energy(BlockedTensor H, BlockedTensor F, Blocke
...
+    size_t ncore = core_mos_.size();
     double E = Efrzc_ + Enuc_;
 
-    for (size_t m = 0, nc = core_mos_.size(); m < nc; ++m) {
-        E += H.block("cc").data()[m * nc + m];
-        E += F.block("cc").data()[m * nc + m];
+#pragma omp parallel for reduction(+ : E)
+    for (size_t m = 0; m < ncore; ++m) {
+        E += H.block("cc").data()[m * ncore + m];
+        E += F.block("cc").data()[m * ncore + m];
```
fevangelista, line 373: Here I don't think that omp really helps us since we pick at most < 10000 elements. What makes sense to avoid is getting the pointer to the data via the data() function, which could be a significant overhead. So I would do instead:
```suggestion
    auto &H_cc = H.block("cc").data();
    auto &F_cc = F.block("cc").data();    
    for (size_t m = 0; m < ncore; ++m) {
        E += H_cc[m * ncore + m];
        E += F_cc[m * ncore + m];
```

fevangelista (approved):

fevangelista:
This is a nice PR! thanks for making these changes. I am running all tests. When they are done and you give me the OK I'll merge it in.

fevangelista:
OK, tests pass. I'll merge after you have a chance to take a look at the points above.

src/mrdsrg-spin-adapted/sa_mrpt2.cc
lcyyork: You are right. Previously, `Vsub` is arranged as `{na, nv, na}` but now is `{na, na, nv, size}`: a switch between second (`i[1]`) and third (`i[2]`) indices.

src/mrdsrg-spin-adapted/sadsrg.cc
lcyyork: This is a good point. I have added a few lines for printing.

src/mrdsrg-spin-adapted/sadsrg.cc
lcyyork: Thank you for the suggestion. I have made the changes.

lcyyork:
The standard test set (158 tests) gives a pass on my computer.

PR #215, opened 2020-12-08
Reset forte options

lcyyork:
## Description
Add a function to reset Forte options. Not sure where this file should live. I tried to put it to utils but that requires py3js installed, which is probably not ideal.

## User Notes
- [x] Reset Forte options in the Python side.

## Checklist
- [ ] Ready to go!

### DIFF
--- a/__init__.py
+++ b/__init__.py
@@ -31,24 +31,13 @@
 __version__ = '1.0'
 __author__ = 'Forte Developers'
 
-import sys
-
 # Load Python modules
 from .pymodule import *
-
-from .register_forte_options import *
 
 # Load C++ plugin
 from .forte import *
 
-# Create a ForteOptions object (stores all options)
-forte_options = forte.ForteOptions()
+from .options_helpers import reset_forte_options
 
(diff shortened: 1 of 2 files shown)


lcyyork:
@fevangelista This may not worth the effort. If so, I will close this PR and delete the branch.

huangm29 (approved):

fevangelista:
@lcyyork : do you want to fix the issue mentioned in the TODO before committing this PR?

lcyyork:
@fevangelista So far the code does not work as intended. There are two ways to continue, as mentioned in the previous GibHub comment.

**This is a design issue.** How do we want the user handle multiple blocks of Forte options and `energy('forte')` calls? I do not know when we need to reset _only_ the Forte options. The good thing is that every `energy('forte')` will start from a new fresh default options. BUT doing so will make the input longer (e.g., I do not want to change `e_convergence` for different forte energy calls).

It seems we are moving towards the direction to eliminate multiple `energy('forte')` calls but more specific functionality-based procedure. The merit of this PR will be further weakened.

PR #219, opened 2020-12-14
Apply orbital orthonormality test to orbitals read from file

lcyyork:
## Description
- This PR considers the orbitals read from a file more carefully, where we also test the orbital orthonormality now.
- This PR also add the functionality of saving energies of all reference relaxation steps in DSRG to a json file (`dsrg_relaxed_energies.json`).
- This PR changes the **default DSRG flow parameter to 0.5**.
- Bug fix on the cumulative sigma threshold in AVAS where the last orbital that go beyond the threshold was not included previously.
- Minor printing improvement in sq-LDSRG(2) where temporary tensors were treated as global objects for memory counting.

## TODO
- [ ] The functions to set Lagrangian matrix in `casscf_orb_grad_deriv.cc` and `casscf_gradient.cc` should be changed accordingly after PR [#2066](https://github.com/psi4/psi4/pull/2066) of Psi4 merged.

## User Notes
- [x] Apply orbital orthonormality test to orbitals read from file.
- [x] Add the keyword `DSRG_DUMP_RELAXED_ENERGIES`, which enables the user to save energies of every reference relaxation step in DSRG.
- [x] **Change the default value of `DSRG_S` to 0.5**.
- [x] Move functions `orbital_projection`, `read_orbitals`, and `dump_orbitals` to `proc/orbital_helpers.py`.

## Checklist
- [x] Documented source code
- [x] Ready to review
- [x] Ready to go!


docs/sphinx/methods/dsrg.rst
```diff
@@ -791,6 +792,13 @@ Max macro iterations for MR-DSRG reference relaxation.
 * Type: integer
 * Default: 15
 
+**DSRG_DUMP_RELAXED_ENERGIES**
+
+Save energies that have been pushed to Psi4 environment for all reference relaxation steps.
```
JonathonMisiewicz, line 797: I didn't understand this description on first reading. Change to "Dump the energies after each reference relaxation step to JSON?"

proc/dsrg.py
```diff
@@ -78,6 +109,7 @@ def __init__(self, active_space_solver, state_weights_map, mo_space_info, ints,
         self.Heff_implemented = False
         self.converged = False
         self.energies = []  # energies along the relaxation steps
+        self.energies_environment = {}  # energies pushed to Psi4 environment
```
JonathonMisiewicz, line 112: Can you change the comment to "Psi4 environment **globals**"? That extra word makes the comment much clearer to me.

pymodule.py
```diff
@@ -38,101 +38,91 @@
...
-    mo_space_info : the MO space info read from options
-        A Forte MOSpaceInfo object
 
-    Returns
-    -------
-    tuple(ForteIntegrals, SCFInfo, MOSpaceInfo)
-        a tuple containing the ForteIntegrals, SCFInfo, and MOSpaceInfo objects
+def check_MO_overlap(options, molecule, Ca):
     """
+    Test the MO overlp matrix is identity or not.
```
JonathonMisiewicz, line 77: overl**a**p

pymodule.py
```diff
@@ -38,101 +38,91 @@
...
+                                           basis_rel, "DECON", basis,
+                                           puream=puream)
+        wfn.set_basisset('BASIS_RELATIVISTIC', rel_bas)
+
+    p4print("\n  Checking orbital orthonormality against current geometry ...")
+
+    # build MO overlap
+    mints = wfn.mintshelper()
+    S = mints.so_overlap()
+    mo_overlap = psi4.core.triplet(Ca, S, Ca, True, False, False)
```
JonathonMisiewicz, line 107: `psi4.core.transform(S, Ca)` will do the same transformation with a little bit less code.

pymodule.py
```diff
@@ -38,101 +38,91 @@
...
+    :param Ca: a Psi4 Matrix that holds orbital coefficients
 
-    # Build Forte SCFInfo object
-    scf_info = forte.SCFInfo(wfn)
-
-    # Build a map from Forte StateInfo to the weights
-    state_weights_map = forte.make_state_weights_map(options, mo_space_info)
-
-    return (state_weights_map, mo_space_info, scf_info)
+    :return: a tuple of (S_MO == I, a new Psi4 Wavefunction object)
```
JonathonMisiewicz, line 86: Why are we building a new Psi wavefunction object in our orthogonality checker?

pymodule.py
```diff
@@ -38,101 +38,91 @@
...
+
+    p4print("\n  Checking orbital orthonormality against current geometry ...")
+
+    # build MO overlap
+    mints = wfn.mintshelper()
+    S = mints.so_overlap()
+    mo_overlap = psi4.core.triplet(Ca, S, Ca, True, False, False)
+
+    # test orbital orthonormality
+    mo_overlap.zero_diagonal()
```
JonathonMisiewicz, line 110: This checks orthogonality, not orthonormality. Checking orthogonality should be enough for the orbital reading errors we're worried about, but it's still different from orthonormality.

src/integrals/diskdf_integrals.cc
```diff
@@ -510,7 +510,8 @@ void DISKDFIntegrals::gather_integrals() {
         }
     }
     df_->set_memory(static_cast<size_t>(mem));
-    df_->set_method("DIRECT");
```
JonathonMisiewicz, line 513: Could you elaborate on why you commented this line?

JonathonMisiewicz (commented):
I started reviewing before I saw the message that "requested reviews" meant "thought you should be aware of this PR." Oh well.

fevangelista (commented):
Thank you @lcyyork for this commit.

src/integrals/diskdf_integrals.cc
lcyyork: Yes, I am actually testing it now (why it has not been marked as Ready to go).

lcyyork:
@JonathonMisiewicz 

> I started reviewing before I saw the message that "requested reviews" meant "thought you should be aware of this PR." Oh well.

Thank you for doing the review.

src/integrals/diskdf_integrals.cc
```diff
@@ -510,7 +510,8 @@ void DISKDFIntegrals::gather_integrals() {
...
+    //    df_->set_method("DIRECT");
```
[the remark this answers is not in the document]
lcyyork: I am still testing here.

register_forte_options.py
```diff
@@ -90,7 +90,8 @@ def register_driver_options(options):
 
     options.add_int("PRINT", 1, "Set the print level.")
 
-    options.add_bool("READ_ORBITALS", True, "Read orbitals from file if true")
+    options.add_bool("READ_ORBITALS", False, "Read orbitals from file if true")
```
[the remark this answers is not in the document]
lcyyork: Yes, it is in conflict with the QC branch, which is done in C++ side. The Python implementation uses PSI API and numpy. Afterall, Ca is a Matrix from Psi4.

I turned it off by default is because reading orbitals now occurs at the very beginning and it will print 'file NOT found' if 'forte_Ca.npz' file is not available. I just don't want this extra line in the output.

pymodule.py
lcyyork: You are right, we do not need that. Let me try to directly create a MintsHelper. My previous intent is to always create a new wave function in Forte. But later I feel like it may not be a good idea for finite difference calculations.

pymodule.py
lcyyork: You are correct, let me change the wording.

pymodule.py
lcyyork: Is it `S.transform(Ca)`? I do not think `transform` is defined (`module 'psi4.core' has no attribute 'transform'`).

pymodule.py
lcyyork: OK, I think the `std::map<std::string, std::shared_ptr<BasisSet>> basissets_` is not exposed for `MintsHelper`.

pymodule.py
lcyyork: OK, I decided to change the function instead.

pymodule.py
fevangelista: In C++ it is what you wrote @lcyyork

pymodule.py
fevangelista: We should make a PR for this in psi4. I am not sure that X2C modifies the overlap matrix, so maybe we do not need it. But psi4 might automatically try to run X2C (even if not needed) and run into trouble.

lcyyork: Yes, maybe in a future PR of fixing the X2C large 1-norm issue.

pymodule.py
lcyyork: I have changed to `S.transform(Ca)  # S = Ca^T S Ca`, but personally I think the `triplet` is more transparent.

JonathonMisiewicz (approved):

PR #243, opened 2021-05-21
Fix bug with repeated operators. Add Tutorial_01.03. Test exception.

fevangelista:
## Description
Fix a bug in the `SQOperator` class that allowed the user to input expressions with repeated operators.

## User Notes
- [x] Throw an exception if `SQOperator` is initialized with repeated operators.
- [x] Added a Tutorial (01.03)

## Checklist
- [x] Documented source code
- [x] Ready to go!


JonathonMisiewicz (approved):
Thanks for the speedy fix and tests! Merge when you're ready.

PR #249, opened 2021-06-10
make setup.py more flexible

lcyyork:
## Description
I just realized there is a setup.py. This PR makes the cmake commands more configurable from setup.cfg. For example,

```
[CMakeBuild]
ambitpath=/Users/york/src/ambit-bin-debug
cmake_build_options=-j8 -v
cmake_config_options=-DCMAKE_C_COMPILER=/usr/local/Cellar/llvm/12.0.0/bin/clang -DCMAKE_CXX_COMPILER=/usr/local/Cellar/llvm/12.0.0/bin/clang++ -DCMAKE_Fortran_COMPILER=/usr/local/Cellar/gcc/10.2.0_4/bin/gfortran
```

## User Notes
- [x] The default maximum number of concurrent processes to use for cmake building is adapted to the number of physical CPU cores (+1 if hyperthreading).

## Checklist
- [x] Ready to go!

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,6 +1,7 @@
 from setuptools import find_packages
 
 import os
+import psutil
 import re
 import sys
 import sysconfig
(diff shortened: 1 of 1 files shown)


lcyyork:
There is no instruction of setting the PYTHONPATH if using the second approach of compiling forte.

fevangelista (approved):
Thanks for pushing this. I'll take care of PYTHONPATH in PR #250.

PR #250, opened 2021-06-10
Add safeguards to get_alfa_occ and friends. Fix issue in setup.py and…

fevangelista:
## User Notes
- [x] Add safeguards to get_alfa_occ function and friends in the Determinant class.
- [x] Fix issues in setup.py
- [x] Document how to setup PYTHONPATH correctly if compiling with CMake

New features will be documented in another PR (#246) once this is merged in.

## Checklist
- [x] Documented source code
- [x] Ready to go!

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,6 +79,11 @@ Forte may also be compiled by directly invoking CMake by following these instruc
   make
  ```
 
+### Setting up the `PYTHONPATH`
+
+If Forte is compiled with CMake, you will need to specify `PYTHONPATH` environment variable to make sure that it can be imported in python. Assuming that Forte is installed in the folder `<fortedir>`, then `PYTHONPATH` should contain `<fortedir>`. Note that if you instead include `<fortedir>/forte` in `PYTHONPATH` you will get an error.
+
+
 The following script automates steps 1 and 2 of the forte compilation process
 
 ```
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@
 [![Documentation Status](https://readthedocs.org/projects/forte/badge/?version=latest)](http://forte.readthedocs.io/en/latest/?badge=latest)
 [![codecov](https://codecov.io/gh/evangelistalab/forte/branch/master/graph/badge.svg)](https://codecov.io/gh/evangelistalab/forte)
 
-#### Code Authors: Evangelistalab
+#### Code Authors: EvangelistaLab
 
 #### Manual (http://forte.readthedocs.io/)
 #### Web: evangelistalab.org
@@ -12,7 +12,7 @@
 Forte is an open-source plugin to Psi4 (https://github.com/psi4/psi4) that implements a variety of quantum chemistry methods
 for strongly correlated electrons.
 
-## Download and compilation
+## Download and compilation of Forte
 
 Prior to the compilation of Forte you must first check to make sure you have the following:
 
@@ -25,7 +25,7 @@ Prior to the compilation of Forte you must first check to make sure you have the
 
 Once you have the current versions of Psi4, CMake, and Ambit, follow the following instructions to install Forte.
 
-### Download
+### Download Forte
 
 1. Open a terminal and change the current working directory to the location where you want to clone the Forte directory.
 Let's assume this is the folder `src`.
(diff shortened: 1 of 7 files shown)


setup.py
```diff
@@ -13,29 +10,36 @@
...
         self.cmake_config_options = ''
-        ncore_phys = psutil.cpu_count(logical=False)
-        ncore_virt = psutil.cpu_count(logical=True)
-        nprocs = ncore_phys + 1 if ncore_phys != ncore_virt else ncore_phys
+        # get the CPU count. None is returned if the number of CPUs is undeterminded.
+        cpu_count = os.cpu_count()
+        if cpu_count is None:
+            nprocs = 1
+        # choose a reasonable number of processes
+        nprocs = cpu_count / 2 + 1 if cpu_count > 4 else cpu_count
```
JonathonMisiewicz, line 42: We may want to allow the user to pass this as a command line argument and use this logic if the argument isn't passed. This may well be outside the scope of the PR.

JonathonMisiewicz (commented):
Mostly LGTM, but I have a few suggestions.

setup.py
lcyyork: It is possible. The `cmake-build-options` should override this. For example, `make -j2 -j8` reads the last one.

forte/api/sci_api.cc
```diff
@@ -57,6 +57,11 @@ void export_Determinant(py::module& m) {
         .def(py::init<const std::vector<bool>&, const std::vector<bool>&>())
         .def("get_alfa_bits", &Determinant::get_alfa_bits, "Get alpha bits")
         .def("get_beta_bits", &Determinant::get_beta_bits, "Get beta bits")
+        .def("get_alfa_occ", &Determinant::get_alfa_occ, "Get a vector of alpha occupied orbitals")
```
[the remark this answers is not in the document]
lcyyork: I kind of agree. We could say "Get the vector of alpha occupied orbital indices".

lcyyork (approved):
I would probably make the api functions a bit more transparent as Jonathon suggested.

README.md
```diff
@@ -79,12 +79,13 @@ Forte may also be compiled by directly invoking CMake by following these instruc
   make
  ```
 
-### Setting up the `PYTHONPATH`
+#### Setting up the `PYTHONPATH`
 
 If Forte is compiled with CMake, you will need to specify `PYTHONPATH` environment variable to make sure that it can be imported in python. Assuming that Forte is installed in the folder `<fortedir>`, then `PYTHONPATH` should contain `<fortedir>`. Note that if you instead include `<fortedir>/forte` in `PYTHONPATH` you will get an error.
```
lcyyork, line 84: I am wondering if we should be more specific on `<fortedir>`. For example, I am now at folder `src` and I `git clone`. I will then have `src/forte`. Then I `cd forte` and do the `cmake` and `make`. Now the `__init__.py` is actually in `src/forte/forte`. So `PYTHONPATH` should include `src/forte` but not `src`.

fevangelista: Good point. I made the entire procedure more explicit about folders.

setup.py
fevangelista: Here I decided that we should go with the maximum number and let the use set a particular value if needed. I added the variable nprocs that can be specified in the setup.cfg file.

forte/sparse_ci/determinant.hpp
```diff
@@ -171,6 +171,12 @@ template <size_t N> class DeterminantImpl : public BitArray<N> {
     /// Return a vector of occupied alpha orbitals
     std::vector<int> get_alfa_occ(int norb) const {
         std::vector<int> occ;
+        if (norb > nbits_half) {
+            throw std::range_error(
+                "Determinant::get_alfa_occ(int norb) was passed a value of norb (" +
+                std::to_string(norb) + ") larger than the maximum number of alpha orbitals (" +
+                std::to_string(nbits_half) + ").");
```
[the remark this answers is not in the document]
fevangelista: Added unit test checks.

forte/api/sci_api.cc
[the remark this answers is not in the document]
fevangelista: Thanks for this point. In forte we consistently use irrep-based ordering of the MOs, so I am not sure this needs to specified (also it depends on how the user decides to fill up the determinant. The ordering matters only once the class is paired to a set of integrals).

fevangelista:
@JonathonMisiewicz , @lcyyork : I think I addressed all of your comments. Take a look and go ahead and merge if OK.

PR #251, opened 2021-06-11
New python interface: HF

fevangelista:
### The point of this PR
This PR implements a high-level Python layer that allows to write a computation using a object-oriented approach (very close to the one used in QForte). For example, a HF computation can be invoked with
```python

from forte.solvers import solver_factory, HF

# define a molecule
xyz = """
H 0.0 0.0 0.0
H 0.0 0.0 1.0
"""

# create a molecular model
base = solver_factory(molecule=xyz, basis='cc-pVDZ')

# specify the electronic state
state = base.state(charge=0, multiplicity=1, sym='ag')

# create a HF class, pass it the base solver and then run
hf = HF(base, state=state)
hf.run()

assert hf.value('hf energy') == pytest.approx(ref_energy, 1.0e-10)
```

### Principles
Some of the current principles
- Replace complex computational workflows with an approach based on composition of solver objects
- Solver classes should be responsible for selecting sensible defaults and check for errors
- Solver classes should not make assumptions about the details of a computation (e.g. spin state, charge, symmetry)
- Don't worry about breaking conventions (e.g. options names, etc.). 

### What has been implemented
Several base classes and the HF solver. Documentation will come after some iterations.
- `Molecule`
- `Basis`
- `Model`: (aka Hamiltonian) responsible for getting information about the system and the Hamiltonian
- `MolecularModel(Model)`: specialization of `Model` for molecules (based on psi4)
- `Solver`: the base class for all computational methods. Holds a data object that handles C++ forte objects (MOSpaceInfo, SCFInfo, ...)
- `BasicSolver(Solver)`: a vanilla specialization of `Solver`. The root of the computational graph.
- `HF(Solver)`: specialization of `Solver` for HF (RHF, ROHF, and UHF). This class takes a `StateInfo` object and will check that the final state has the correct symmetry.

## User Notes
- [x] Adds Basis, Molecule, Model, MolecularModel classes
- [x] Adds HF solver and tests (in `tests/pytest/hf`)
- [x] Adds sphinx docs for HF and Molecule classes 
- [x] Adds tests for the HF solver (RHF,UHF)
- [x] Cleaned the Symmetry class
- [x] Adds the C++ function `find_case_insensitive`, which searches a `string` in a `vector<string>` in a case insensitive way.
- [x] `forte.startup()` is now invoked when calling `import forte`. So there is no need for it in scripts. `forte.cleanup()` should also be removed (although it is never called)

## Checklist
- [x] Documented source code
- [x] Ready to go!


JonathonMisiewicz:
A philosophical question that kept coming up during the code review: when do we want to have tests and documentation?

Example: Methods of the `Basis` class have unit tests and good documentation. But the newly created and exposed `SCFInfo::nmopi` has neither. I'm familiar enough with Psi conventions to figure it out, but that is not true of most developers. How do we know to test and document one but not the other?

fevangelista:
You have a good point. Unfortunately several of the C++ Forte classes are not documented/tested on the python side. I am trying to be more consistent in documenting/testing the new code I am pushing and I'd be happy to add more docs as we add more code to the older classes. But this is a place where certainly all contributing developers (@lcyyork , @jeffschriber , @tyzhang1993 , @henankf223 , @huangm29 , @ustcwsh) can help. Especially just by adding docstrings to the pybind11 interface.

JonathonMisiewicz:
Okay, so the standard is "document and unit-test new code that's also Python-exposed," and getting the rest is a longer term project. That, I can work with, thanks.

forte/helpers/symmetry.cc
```diff
@@ -0,0 +1,80 @@
...
+
+const std::string& Symmetry::point_group_label() const { return pg_; }
+
+const std::vector<std::string>& Symmetry::irrep_labels() const {
+    return __pg_to_irrep_labels.at(pg_);
+}
+
+const std::string& Symmetry::irrep_label(size_t h) const {
+    if (h >= nirrep()) {
+        std::string msg = "(Symmetry) irrep " + std::to_string(h) +
```
JonathonMisiewicz, line 57: +1 for the error catching and message

forte/helpers/symmetry.cc
```diff
@@ -0,0 +1,80 @@
...
+#include "helpers/symmetry.h"
+
+namespace forte {
+
+Symmetry::Symmetry(std::string point_group) {
+    to_upper_string(point_group); // capitalize
+    // check if this is a valid point group label
+    auto search = std::find(__point_groups.cbegin(), __point_groups.cend(), point_group);
+    if (search == __point_groups.end()) {
+        std::string msg = "Point group " + point_group + " not found or supported by forte.\n";
```
JonathonMisiewicz, line 43: Codecoverage has a point. It may be worht doing a test that the error catching works.

forte/model.py
```diff
@@ -0,0 +1,109 @@
+from abc import ABC, abstractmethod
+from forte.molecule import Molecule
+from forte.basis import Basis
+from forte.forte import StateInfo
+from forte.forte import Symmetry
+
+
+class Model(ABC):
```
JonathonMisiewicz, line 8: It'd be worth saying what a model is. I take it this is a model of a system, with a Hamiltonian?

forte/model.py
```diff
@@ -0,0 +1,109 @@
...
+        self._molecule = molecule
+        self._basis = basis
+        self._scf_aux_basis = scf_aux_basis
+        self._corr_aux_basis = corr_aux_basis
+        self.symmetry = Symmetry(molecule.molecule.point_group().symbol().capitalize())
+
+    def __repr__(self):
+        """
+        return a string representation of this object
+        """
```
JonathonMisiewicz, line 39: Do you want to implement this?

forte/model.py
```diff
@@ -0,0 +1,109 @@
...
+
+    def __str__(self):
+        """
+        return a string representation of this object
```
JonathonMisiewicz, line 43: Do you want to implement this?

forte/model.py
```diff
@@ -0,0 +1,109 @@
...
+        if self._corr_aux_basis is None:
+            return None
+        return self._corr_aux_basis.basis
+
+    @property
+    def point_group(self) -> str:
+        return self.symmetry.point_group_label()
+
+    def state(self, charge: int, multiplicity: int, ms: float = None, sym: str = None):
+        """This function is used to create a StateInfo object. It checks for potential errors."""
```
JonathonMisiewicz, line 71: The meaning of these arguments should be documented.

forte/results.py
```diff
@@ -0,0 +1,20 @@
+class Results:
+    """
+    A class used to store the results of a computation.
+    """
+    def __init__(self):
+        self._data = dict()
+
+    def add(self, label, value, description, units):
```
JonathonMisiewicz, line 8: Is there a way for a user to access a description?

forte/results.py
```diff
@@ -0,0 +1,20 @@
+class Results:
```
JonathonMisiewicz, line 1: I don't think we have a unit test file for this.

tests/pytest/api/reference_data/casscf.out
```diff
@@ -0,0 +1,11 @@
+molecule {
```
JonathonMisiewicz, line 1: Are you sure this should be .out?

tests/pytest/hf/test_uhf.py
```diff
@@ -0,0 +1,66 @@
...
+    root = solver_factory(molecule=mol, basis=basis)
+
+    # specify the electronic state
+    state = root.state(charge=0, multiplicity=3, sym='a1')
+
+    # define a HF object
+    hf = HF(root, state=state, docc=[2, 0, 0, 1], socc=[1, 0, 1, 0], restricted=False)
+
+    # should raise an error since we get a state with different symmetry
+    with pytest.raises(RuntimeError):
```
JonathonMisiewicz, line 60: Do we want to run validation when the HF object is created, or when you call its `run` method? The inconsistency with the occupation and the target state is clear from the arguments.

forte/solvers/solver.py
```diff
@@ -0,0 +1,119 @@
...
+
+    def make_mo_space_info(self, mo_spaces):
+        nmopi = self.scf_info.nmopi()
+        return forte.make_mo_space_info_from_map(nmopi, self.model.point_group, mo_spaces)
+
+    def state(self, charge: int, multiplicity: int, ms: float = None, sym: str = None):
+        return self.data.model.state(charge, multiplicity, ms, sym)
+
+
+class BasicSolver(Solver):
```
JonathonMisiewicz, line 89: What's the purpose of this class?

forte/solvers/callback_handler.py
```diff
@@ -0,0 +1,45 @@
+class CallbackHandler():
+    """
+    This class stores a list of callback functions labeled by an ID.
```
JonathonMisiewicz, line 3: This isn't a list. This is a dict.

tests/pytest/hf/test_rhf.py
```diff
@@ -0,0 +1,91 @@
...
+    # create a basis object
+    basis = Basis('cc-pVDZ')
+
+    # create a molecular model
+    root = solver_factory(molecule=mol, basis=basis)
+
+    # specify the electronic state
+    state = root.state(charge=0, multiplicity=1, sym='a1')
+
+    ch = CallbackHandler()
```
JonathonMisiewicz, line 78: Please add an example where the CallbackHandler is used to _do something_.

JonathonMisiewicz (commented):
This is great overall, but I do have a few things that should be added, largely examples, tests, and documentation.

For future refernece, I'd have preferred if file formatting and messing with setup was moved to separate PR.

forte/__init__.py
```diff
@@ -46,6 +56,9 @@
 # Register options defined in Forte in the forte_options object
 register_forte_options(forte_options)
 
+# create a singleton to handle startup and cleanup of forte
+forte_manager = ForteManager()
```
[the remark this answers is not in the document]
fevangelista: You are right. Removed `forte_manager = ` here.

forte/api/forte_python_module.cc
```diff
@@ -123,6 +124,18 @@ void export_MCSCF_2STEP(py::module& m) {
         .def("compute_energy", &MCSCF_2STEP::compute_energy, "Compute the MCSCF energy");
 }
 
+void export_Symmetry(py::module& m) {
+    py::class_<Symmetry>(m, "Symmetry")
+        .def(py::init<std::string>())
+        .def("point_group_label", &Symmetry::point_group_label, "the label of this point group")
+        .def("irrep_labels", &Symmetry::irrep_labels, "vector of irrep labels")
+        .def("irrep_label", &Symmetry::irrep_label, "the label of irrep h")
```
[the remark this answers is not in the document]
fevangelista: When I run sphinx and ask to document this class I do get the function signature (see figure). So I think a one-line comment should be enough. Do you have any suggestion for a  standard function docstring format? I know we are sloppy with ending the description with periods.
image

forte/api/forte_python_module.cc
```diff
@@ -170,6 +182,10 @@ PYBIND11_MODULE(forte, m) {
           "Make a DSRG pointer (spin-adapted implementation)");
     m.def("make_casscf", &make_casscf, "Make a CASSCF object");
     m.def("make_mcscf_two_step", &make_mcscf_two_step, "Make a 2-step MCSCF object");
+    m.def("make_mcscf",
```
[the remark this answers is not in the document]
fevangelista: This function was leftover from some tests I was doing. Removed.

forte/api/mospaceinfo_api.cc
```diff
@@ -70,7 +70,9 @@ void export_MOSpaceInfo(py::module& m) {
         .def("space_names", &MOSpaceInfo::space_names, "Return the names of orbital spaces")
         .def("irrep_labels", &MOSpaceInfo::irrep_labels, "Return the labels of the irreps")
         .def("irrep_label", &MOSpaceInfo::irrep_label, "Return the labels of the irreps")
-        .def("point_group_label", &MOSpaceInfo::point_group_label, "Return the labels of the irreps");
+        .def("point_group_label", &MOSpaceInfo::point_group_label,
```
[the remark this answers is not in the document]
fevangelista: The former. Fixed.

tests/pytest/api/reference_data/casscf.out
fevangelista: Should be .inp, fixed.

tests/pytest/hf/test_uhf.py
fevangelista: This is a good question. The symmetry cannot be checked at creation, since without a docc/socc, the psi4 HF algorithm is allowed to identify the best solution (but is not aware of a target symmetry) and could end up choosing a state with the wrong symmetry. Ideally, one day we could write a simple HF code that can converge on a desired symmetry. An alternative would be to ask the user to always specify the docc/socc, but that would still make the HF computation prone to error. Unfortunately no easy answer here!

forte/solvers/solver.py
fevangelista: In the current workflow, this is the equivalent to an empty `Wavefunction` object in psi4. It is used later in the code in this way:
```python
    # create an empty solver and pass the model in
    solver = BasicSolver()
    solver.data.model = MolecularModel(
        molecule=molecule, basis=basis, scf_aux_basis=scf_aux_basis, corr_aux_basis=corr_aux_basis
    )
```

forte/solvers/callback_handler.py
fevangelista: Correct.

tests/pytest/hf/test_rhf.py
fevangelista: Ok, this is a bit more challenging. Let me do some coding.

forte/api/forte_python_module.cc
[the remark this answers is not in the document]
JonathonMisiewicz: Oh, I didn't realize that! Very good. The two guidelines I think are worth following are:
* always have a _verb_ that explains what the function does. So in this case, `returns`
* if we use a variable's name in the docstring, have the variable's name be in the pybind
I don't care so much about periods.

forte/solvers/solver.py
JonathonMisiewicz: I forgot that you can't instantiate an ABC. This makes sense.

tests/pytest/hf/test_rhf.py
JonathonMisiewicz: Nice example!

docs/source/api/basis.rst
```diff
@@ -0,0 +1,6 @@
+Basis set
```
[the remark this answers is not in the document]
fevangelista: We use readthedocs, but it's a bit cumbersome to setup doc builds for every new pr.

forte/__init__.py
```diff
@@ -30,14 +30,24 @@
...
+from .basis import Basis
+from .results import Results
+from .model import Model, MolecularModel
+
 from .pymodule import *
 from .register_forte_options import *
-from .core import *
+from .core import ForteManager, clean_options
 from .forte import *
+from .active_space import *
```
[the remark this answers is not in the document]
fevangelista: Correct. Now deleted.

forte/base_classes/mo_space_info.cc
```diff
@@ -49,36 +51,49 @@ MOSpaceInfo::MOSpaceInfo(const psi::Dimension& nmopi, const std::string& point_g
     }
 }
 
-size_t MOSpaceInfo::size(const std::string& space) {
+std::string MOSpaceInfo::str() const {
```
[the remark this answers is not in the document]
fevangelista: Added to unit test.

forte/base_classes/mo_space_info.cc
```diff
@@ -173,7 +188,8 @@ std::vector<std::pair<size_t, size_t>> MOSpaceInfo::relative_mo(const std::strin
     return result;
 }
 
-std::vector<std::pair<size_t, size_t>> MOSpaceInfo::get_relative_mo(const std::string& space) {
+std::vector<std::pair<size_t, size_t>>
+MOSpaceInfo::get_relative_mo(const std::string& space) const {
```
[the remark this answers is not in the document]
fevangelista: `get_` was deprecated just in case someone was still using the old function name. Removed.

forte/basis.py
```diff
@@ -0,0 +1,33 @@
...
+        return f'Basis(\'{self._basis}\')'
+
+    def __str__(self):
+        """
+        return a string representation of this object
+        """
+        return self._basis
+
+    @property
+    def basis(self):
```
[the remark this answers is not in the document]
fevangelista: Now removed.

forte/helpers/symmetry.cc
fevangelista: Added.

forte/model.py
fevangelista: I added a tentative description.

forte/model.py
fevangelista: Implemented.

forte/model.py
fevangelista: Implemented.

forte/model.py
fevangelista: Documented.

forte/results.py
fevangelista: Added access to description and units.

forte/results.py
fevangelista: Added.

JonathonMisiewicz (approved):
It took a bit, but LGTM. Unless you want me to hold off, I can merge once test suite passes.

fevangelista:
Yes, please merge once the tests pass. Thanks again for the suggestions. The next item on the list is FCI (and maybe active space solvers) in general.

PR #254, opened 2021-06-16
DSRG-MRPT2 Gradient 

ustcwsh:
## Description

## User Notes
- [x] Features added

## Checklist
- [x] Documented source code
- [ ] Ready to go!


forte/api/rdms_api.cc
```diff
@@ -92,5 +92,6 @@ void export_RDMs(py::module& m) {
         .def(
             "L3bbb", [](RDMs& rdm) { return ambit_to_np(rdm.L3bbb()); },
             "Return the beta-beta-beta 3-cumulant as a numpy array");
+
```
lcyyork, line 95:
```suggestion
```

forte/base_classes/active_space_method.h
```diff
@@ -133,6 +134,17 @@ class ActiveSpaceMethod {
     /// @param options the options passed in
     virtual void set_options(std::shared_ptr<ForteOptions> options) = 0;
 
+    /// Return the coupling coefficients in ambit::Tensor format
+    virtual CouplingCoefficients coupling_coefficients(int level) {
+        throw std::runtime_error("Not Implemented other than FCI_MO! level = " +
```
lcyyork, line 139: Make sure you run clang format for all your files.

forte/base_classes/active_space_solver.h
```diff
@@ -130,6 +131,12 @@ class ActiveSpaceSolver {
     /// Set if read wave function from file as initial guess
     void set_read_initial_guess(bool read_guess) { read_initial_guess_ = read_guess; }
 
+    /// Return the coupling coefficients for a given state
+    CouplingCoefficients coupling_coefficients(const StateInfo& state, int max_level = 2);
```
lcyyork, line 135:
```suggestion
    CouplingCoefficients coupling_coefficients(const StateInfo& state, int max_level);
```

forte/mrdsrg-spin-integrated/active_dsrgpt2.cc
```diff
@@ -629,6 +629,14 @@ void ACTIVE_DSRGPT2::compute_osc_ref(
     }
 }
 
+
```
lcyyork, line 632: Please remove lines 632 to 639. This file will be moved to attic.

forte/mrdsrg-helper/dsrg_source.h
```diff
@@ -50,6 +50,29 @@ class DSRG_SOURCE {
...
+            switch(i) {
+                case 1:
+                    return (1.0 - std::exp(-s_ * D * D)) / D;
+                case 2:
+                    return (1.0 - std::exp(-s_ * D * D)) / D / D;
+            }
+        }
+    }
+
+    double get_s() {
```
lcyyork, line 72: I don't think this is useful because s is set from outside. But if you insist to keep it, please change to
```suggestion
    double s() {
```

forte/mrdsrg-helper/dsrg_source.h
```diff
@@ -50,6 +50,29 @@ class DSRG_SOURCE {
     /// Renormalize denominator
     virtual double compute_renormalized_denominator(const double& D) = 0;
 
+    double compute_denominator(const double& D, int i) {
```
lcyyork, line 53: This is more or less implemented in `STD_SOURCE`. You should probably create a function here, something like
```
/// Derivative of renormalized denominator with respect to Delta
virtual double compute_renormalized_denominator_deriv(const double& D) {
    throw std::runtime_error("Not implemented. Child class should override this function.");
}
```
And then implement in `STD_SOURCE`.

forte/mrdsrg-spin-integrated/dsrg_mrpt3.cc
```diff
@@ -273,6 +273,13 @@ void DSRG_MRPT3::build_tei(BlockedTensor& V) {
     }
 }
 
+SharedMatrix DSRG_MRPT3::compute_gradient() {
```
lcyyork, line 276: Remove.

forte/mrdsrg-spin-integrated/dsrg_mrpt3.h
```diff
@@ -54,6 +54,9 @@ class DSRG_MRPT3 : public MASTER_DSRG {
     /// Compute the DSRG-MRPT3 energy
     virtual double compute_energy();
 
+    /// Compute gradient
```
lcyyork, line 57: Remove.

forte/mrdsrg-spin-integrated/dwms_mrpt2.cc
```diff
@@ -282,6 +282,13 @@ double DWMS_DSRGPT2::compute_energy() {
     return Ept_[0][0];
 }
 
+SharedMatrix DWMS_DSRGPT2::compute_gradient() {
```
lcyyork, line 285: Remove.

forte/mrdsrg-spin-integrated/dwms_mrpt2.h
```diff
@@ -56,6 +56,9 @@ class DWMS_DSRGPT2 {
     /// compute energy and return the ground state energy
     double compute_energy();
 
+    /// Compute gradient
+    psi::SharedMatrix compute_gradient();
```
lcyyork, line 60: Remove.

forte/mrdsrg-spin-integrated/master_mrdsrg.h
```diff
@@ -38,6 +38,9 @@ class MASTER_DSRG : public DynamicCorrelationSolver {
     /// Compute energy
     virtual double compute_energy() = 0;
 
+    /// Compute the DSRG gradient
+    virtual psi::SharedMatrix compute_gradient() = 0;
```
lcyyork, line 42: This function may not be pure virtual. We could throw a runtime error saying it is only implemented for DSRG-MRPT2.

forte/mrdsrg-spin-integrated/mcsrgpt2_mo.cc
```diff
@@ -222,6 +222,13 @@ double MCSRGPT2_MO::ElementRH(const std::string& source, const double& D, const
     }
 }
 
+SharedMatrix MCSRGPT2_MO::compute_gradient() {
```
lcyyork, line 225: Remove.

forte/mrdsrg-spin-integrated/mcsrgpt2_mo.h
```diff
@@ -69,6 +69,9 @@ class MCSRGPT2_MO {
     /// Compute the energy
     double compute_energy();
 
+    /// Compute gradient
+    psi::SharedMatrix compute_gradient();
```
lcyyork, line 73: Remove.

forte/mrdsrg-spin-integrated/mrdsrg.cc
```diff
@@ -189,6 +189,13 @@ void MRDSRG::print_options() {
                            calculation_info_bool, calculation_info_double, calculation_info_int);
 }
 
+SharedMatrix MRDSRG::compute_gradient() {
```
lcyyork, line 192: Remove.

forte/mrdsrg-spin-integrated/mrdsrg.h
```diff
@@ -61,6 +61,9 @@ class MRDSRG : public MASTER_DSRG {
     /// Compute the corr_level energy with fixed reference
     virtual double compute_energy() override;
 
+    /// Compute gradient
+    psi::SharedMatrix compute_gradient() override;
```
lcyyork, line 65: Remove.

forte/mrdsrg-spin-integrated/three_dsrg_mrpt2.cc
```diff
@@ -335,6 +335,13 @@ void THREE_DSRG_MRPT2::print_options_summary() {
 
 void THREE_DSRG_MRPT2::cleanup() {}
 
+SharedMatrix THREE_DSRG_MRPT2::compute_gradient() {
```
lcyyork, line 338: Remove.

forte/base_classes/coupling_coefficients.h
```diff
@@ -0,0 +1,61 @@
+#ifndef COUPLINGCOEFFICIENTS_H
+#define COUPLINGCOEFFICIENTS_H
+
+#include <ambit/tensor.h>
+
+namespace forte {
+
+class CouplingCoefficients {
```
fevangelista, line 8: I am wondering: what is the purpose of this class? Acting as container for coupling coefficients?
Base classes are important, so they should be documented like all the other ones in the same folder. E.g. see MOSpaceInfo or the ActiveSpace solver. Also I would suggest renaming to `CICouplingCoefficients` since this is specific to CI (e.g., it does not apply to methods like DMRG).

forte/mrdsrg-spin-integrated/three_dsrg_mrpt2.h
```diff
@@ -53,6 +53,9 @@ class THREE_DSRG_MRPT2 : public MASTER_DSRG {
     /// Compute the DSRG-MRPT2 energy
     virtual double compute_energy();
 
+    /// Compute gradient
+    psi::SharedMatrix compute_gradient() override;
```
lcyyork, line 57: Remove.

forte/mrdsrg-spin-integrated/three_dsrg_mrpt2.h
```diff
@@ -53,6 +53,9 @@ class THREE_DSRG_MRPT2 : public MASTER_DSRG {
...
+
```
lcyyork, line 58:
```suggestion
```

lcyyork, line 58:
```suggestion
```

forte/mrdsrg-spin-integrated/three_dsrg_mrpt2.cc
```diff
@@ -335,6 +335,13 @@ void THREE_DSRG_MRPT2::print_options_summary() {
 
 void THREE_DSRG_MRPT2::cleanup() {}
 
+SharedMatrix THREE_DSRG_MRPT2::compute_gradient() {   
+    outfile->Printf("\n Oh no!, you tried to compute gradient but this "
+                    "is not there!!");
+    throw psi::PSIEXCEPTION("gradient is not implemented");
+    return std::make_shared<Matrix>("nullptr", 0, 0);
+}
+
```
lcyyork, line 344:
```suggestion
```

forte/base_classes/active_space_method.h
```diff
@@ -133,6 +134,17 @@ class ActiveSpaceMethod {
...
+                                 std::to_string(level));
+    }
+
+    /// Return the eigen vectors
+    virtual std::vector<ambit::Tensor> eigen_vectors() {
+        throw std::runtime_error("Not Implemented!");
```
fevangelista, line 145: Please throw a more meaningful message and rename function, e.g.
```suggestion
        throw std::runtime_error("ActiveSpaceMethod::eigenvectors(): Not Implemented for this class!");
```

forte/mrdsrg-spin-integrated/mrdsrg.h
```diff
@@ -61,6 +61,9 @@ class MRDSRG : public MASTER_DSRG {
...
+
```
lcyyork, line 66:
```suggestion
```

forte/mrdsrg-spin-integrated/mrdsrg.cc
```diff
@@ -189,6 +189,13 @@ void MRDSRG::print_options() {
                            calculation_info_bool, calculation_info_double, calculation_info_int);
 }
 
+SharedMatrix MRDSRG::compute_gradient() {   
+    outfile->Printf("\n Oh no!, you tried to compute gradient but this "
+                    "is not there!!");
+    throw psi::PSIEXCEPTION("gradient is not implemented");
+    return std::make_shared<Matrix>("nullptr", 0, 0);
+}
+
```
lcyyork, line 198:
```suggestion
```

forte/mrdsrg-spin-integrated/mcsrgpt2_mo.h
```diff
@@ -69,6 +69,9 @@ class MCSRGPT2_MO {
...
+
```
lcyyork, line 74:
```suggestion
```

forte/base_classes/active_space_solver.cc
```diff
@@ -826,4 +827,13 @@ compute_average_state_energy(const std::map<StateInfo, std::vector<double>>& sta
     return average_energy;
 }
 
+CouplingCoefficients ActiveSpaceSolver::coupling_coefficients(const StateInfo& state,
+                                                              int max_level) {
+    return state_method_map_[state]->coupling_coefficients(max_level);
+}
+
+std::vector<ambit::Tensor> ActiveSpaceSolver::eigen_vectors(const StateInfo& state) {
```
fevangelista, line 835:
```suggestion
std::vector<ambit::Tensor> ActiveSpaceSolver::eigenvectors(const StateInfo& state) {
```

forte/mrdsrg-spin-integrated/mcsrgpt2_mo.cc
```diff
@@ -222,6 +222,13 @@ double MCSRGPT2_MO::ElementRH(const std::string& source, const double& D, const
     }
 }
 
+SharedMatrix MCSRGPT2_MO::compute_gradient() {   
+    outfile->Printf("\n Oh no!, you tried to compute gradient but this "
+                    "is not there!!");
+    throw psi::PSIEXCEPTION("gradient is not implemented");
+    return std::make_shared<Matrix>("nullptr", 0, 0);
+}
+
```
lcyyork, line 231:
```suggestion
```

forte/base_classes/active_space_solver.h
```diff
@@ -130,6 +131,12 @@ class ActiveSpaceSolver {
...
+
+    /// Return the eigen vectors for a given state
+    std::vector<ambit::Tensor> eigen_vectors(const StateInfo& state);
```
fevangelista, line 138:
```suggestion
    std::vector<ambit::Tensor> eigenvectors(const StateInfo& state);
```

forte/mrdsrg-spin-integrated/dwms_mrpt2.h
```diff
@@ -56,6 +56,9 @@ class DWMS_DSRGPT2 {
...
+
```
lcyyork, line 61:
```suggestion
```

forte/base_classes/coupling_coefficients.h
```diff
@@ -0,0 +1,61 @@
+#ifndef COUPLINGCOEFFICIENTS_H
+#define COUPLINGCOEFFICIENTS_H
```
fevangelista, line 2:
```suggestion
#define _coupling_coefficients_h_
```

forte/mrdsrg-spin-integrated/dwms_mrpt2.cc
```diff
@@ -282,6 +282,13 @@ double DWMS_DSRGPT2::compute_energy() {
     return Ept_[0][0];
 }
 
+SharedMatrix DWMS_DSRGPT2::compute_gradient() {   
+    outfile->Printf("\n Oh no!, you tried to compute gradient but this "
+                    "is not there!!");
+    throw psi::PSIEXCEPTION("gradient is not implemented");
+    return std::make_shared<Matrix>("nullptr", 0, 0);
+}
+
```
lcyyork, line 291:
```suggestion
```

forte/mrdsrg-spin-integrated/dsrg_mrpt3.h
```diff
@@ -54,6 +54,9 @@ class DSRG_MRPT3 : public MASTER_DSRG {
...
+    psi::SharedMatrix compute_gradient() override;
+
```
lcyyork, line 59:
```suggestion
```

forte/mrdsrg-spin-integrated/dsrg_mrpt3.cc
```diff
@@ -273,6 +273,13 @@ void DSRG_MRPT3::build_tei(BlockedTensor& V) {
     }
 }
 
+SharedMatrix DSRG_MRPT3::compute_gradient() {   
+    outfile->Printf("\n Oh no!, you tried to compute gradient but this "
+                    "is not there!!");
+    throw psi::PSIEXCEPTION("gradient is not implemented");
+    return std::make_shared<Matrix>("nullptr", 0, 0);
+}
+
```
lcyyork, line 282:
```suggestion
```

forte/mrdsrg-spin-integrated/active_dsrgpt2.cc
```diff
@@ -629,6 +629,14 @@ void ACTIVE_DSRGPT2::compute_osc_ref(
...
+SharedMatrix ACTIVE_DSRGPT2::compute_gradient() {   
+    outfile->Printf("\n Oh no!, you tried to compute gradient but this "
+                    "is not there!!");
+    throw psi::PSIEXCEPTION("gradient is not implemented");
+    return std::make_shared<Matrix>("nullptr", 0, 0);
+}
+
```
lcyyork, line 639:
```suggestion
```

forte/mrdsrg-spin-integrated/active_dsrgpt2.h
```diff
@@ -83,6 +83,9 @@ class ACTIVE_DSRGPT2 {
     /// Compute energy
     double compute_energy();
 
+    /// Compute gradient
+    psi::SharedMatrix compute_gradient() override;
+
```
lcyyork, line 88:
```suggestion
```

forte/mrdsrg-helper/dsrg_source.h
```diff
@@ -50,6 +50,29 @@ class DSRG_SOURCE {
...
+        if (std::fabs(D) < 1e-6) {
```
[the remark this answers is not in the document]
fevangelista: Also, I would make another function to handle the modified denominator instead of using `switch` here.

forte/pymodule.py
```diff
@@ -638,9 +657,10 @@ def gradient_forte(name, **kwargs):
 
     # Run a method
     job_type = options.get_str('JOB_TYPE')
+    correlation_solver = options.get_str('CORRELATION_SOLVER')
 
-    if job_type not in {"CASSCF", "MCSCF_TWO_STEP"}:
-        raise Exception('Analytic energy gradients are only implemented for job_types CASSCF and MCSCF_TWO_STEP.')
+    if job_type not in {"CASSCF", "MCSCF_TWO_STEP"} and not correlation_solver == 'DSRG-MRPT2':
+        raise Exception('Analytic energy gradients are only implemented for job_types CASSCF, DSRG-MRPT2 and MCSCF_TWO_STEP.')
```
lcyyork, line 663:
```suggestion
        raise Exception('Analytic energy gradients are only implemented for job_types CASSCF, MCSCF_TWO_STEP, and DSRG-MRPT2.')
```

forte/mrdsrg-spin-integrated/dsrg_mrpt3.cc
fevangelista, line 276: There is another way to do this: implement and make `compute_gradient()` throw an exception in the base class, and then overload it only for the classes for which it is implemented. This extra code in each of the derived classes is difficult to maintain.

forte/pymodule.py
```diff
@@ -674,6 +694,10 @@ def gradient_forte(name, **kwargs):
                                            mo_space_info, ints)
         energy = casscf.compute_energy()
 
+    if correlation_solver == 'DSRG-MRPT2':
```
lcyyork, line 697:
```suggestion
    if job_type == 'NEWDRIVER' and correlation_solver == 'DSRG-MRPT2':
```

tests/methods/dsrg-mrpt2-gradient-1/input.dat
```diff
@@ -0,0 +1,58 @@
...
+
+
+grad = gradient('forte')
+
+compare_matrices(ref_grad, grad, 6, "DSRG-MRPT2 gradient on 4 H atoms with c1 symmetry")
+
+
+
+
+
```
fevangelista, line 58:
```suggestion
```

tests/methods/dsrg-mrpt2-gradient-2/input.dat
```diff
@@ -0,0 +1,52 @@
...
+F 1 0.9
+}
+
+set {
+  basis cc-pvdz
+  reference rhf
+  scf_type pk
+  e_convergence 10
+  d_convergence 10
+  # docc [3,0,1,1]
```
fevangelista, line 23:
```suggestion
```

tests/methods/dsrg-mrpt2-gradient-3/input.dat
```diff
@@ -0,0 +1,50 @@
+#! Generated using commit GITCOMMIT
```
fevangelista, line 1:
```suggestion
```

tests/methods/dsrg-mrpt2-gradient-2/input.dat
```diff
@@ -0,0 +1,52 @@
+#! Generated using commit GITCOMMIT
```
fevangelista, line 1:
```suggestion
```

tests/methods/dsrg-mrpt2-gradient-1/input.dat
```diff
@@ -0,0 +1,58 @@
+#! Generated using commit GITCOMMIT
```
fevangelista, line 1:
```suggestion
```

forte/mrdsrg-helper/dsrg_source.h
ustcwsh: I will check it.

forte/mrdsrg-spin-integrated/dsrg_mrpt2.h
```diff
@@ -175,6 +178,190 @@ class DSRG_MRPT2 : public MASTER_DSRG {
...
+    void set_z();
+    void set_alpha();
+    void set_CI();
+    void set_b();
+    void set_z_cc();
+    void set_z_vv();
+    void set_z_aa_diag();
+    void set_w();
+
+    double s;
```
fevangelista, line 237: Is this needed? We should already have s stored somewhere (`s_`?)

forte/mrdsrg-spin-integrated/dsrg_mrpt2.h
```diff
@@ -175,6 +178,190 @@ class DSRG_MRPT2 : public MASTER_DSRG {
...
+    /// Dimension of different irreps
+    psi::Dimension irrep_vec;
+
+    /// Two-body denisty tensor
+    ambit::BlockedTensor Gamma2_;
+    /// Lagrangian tensor
+    ambit::BlockedTensor W_;
+
+    // core Hamiltonian
+    ambit::BlockedTensor H;
```
fevangelista, line 276: Same as above: don't we already have H, V, F, Gamma2 already available?

tests/methods/dsrg-mrpt2-gradient-1/input.dat
lcyyork, line 58:
```suggestion
```

tests/methods/dsrg-mrpt2-gradient-1/input.dat
lcyyork, line 1: Add description on this input. So for the other test cases.

forte/base_classes/coupling_coefficients.h
fevangelista: There might be a similar change to make at the end of the file

forte/pymodule.py
```diff
@@ -478,10 +478,29 @@ def forte_driver(state_weights_map, scf_info, options, ints, mo_space_info):
     # solver for dynamical correlation from DSRG
     correlation_solver_type = options.get_str('CORRELATION_SOLVER')
     if correlation_solver_type != 'NONE':
+
+        # Grab coupling coefficients
+        # NOTE: Orbitals have to be semicanonicalized already to make sure
+        #       DSRG reads consistent CI coefficients before and after SemiCanonical class.
+        do_grad = options.get_str('DERTYPE') == 'FIRST'
+        if active_space_solver_type == "CAS" and do_grad:
+            # This is OK only when running state-specific calculations
```
lcyyork, line 487:
```suggestion
            # This is OK only when running ground-state calculations
```

forte/pymodule.py
```diff
@@ -478,10 +478,29 @@ def forte_driver(state_weights_map, scf_info, options, ints, mo_space_info):
...
+        if do_grad:
+            dsrg_proc.compute_gradient(coupling_coefficients, ci_vectors)
+        else:
+            return_en = dsrg_proc.compute_energy()
+
         dsrg_proc.print_summary()
         dsrg_proc.push_to_psi4_environment()
+
+        if do_grad:
+            return True
```
lcyyork, line 503:
```suggestion
        return_en = dsrg_proc.compute_energy()
        dsrg_proc.print_summary()
        dsrg_proc.push_to_psi4_environment()

        if do_grad:
            dsrg_proc.compute_gradient(coupling_coefficients, ci_vectors)
```

forte/pymodule.py
```diff
@@ -478,10 +478,29 @@ def forte_driver(state_weights_map, scf_info, options, ints, mo_space_info):
...
+        # Grab coupling coefficients
+        # NOTE: Orbitals have to be semicanonicalized already to make sure
+        #       DSRG reads consistent CI coefficients before and after SemiCanonical class.
+        do_grad = options.get_str('DERTYPE') == 'FIRST'
+        if active_space_solver_type == "CAS" and do_grad:
+            # This is OK only when running state-specific calculations
+            state = list(state_map.keys())[0]
+            coupling_coefficients = active_space_solver.coupling_coefficients(state, 3)
+            ci_vectors = active_space_solver.eigen_vectors(state)
+
```
lcyyork, line 491:
```suggestion
```

forte/proc/dsrg.py
```diff
@@ -245,6 +245,26 @@ def compute_energy(self):
...
+    def compute_gradient(self, coupling_coefficients, ci_vectors):
+        """ Compute DSRG-MRPT2 analytic gradients. """
+        # Notes (Shuhe Wang):
+
+        e_relax = 0.0  # initialize this to avoid PyCharm warning
+
+        # Perform the initial un-relaxed DSRG
+        self.make_dsrg_solver()
+        self.dsrg_setup()
+        e_dsrg = self.dsrg_solver.compute_energy()
```
lcyyork, line 257: I think this function should only compute gradient but not energy. 
```suggestion
```

forte/proc/dsrg.py
```diff
@@ -245,6 +245,26 @@ def compute_energy(self):
...
+
+        self.dsrg_solver.set_coupling_coefficients(coupling_coefficients)
+        self.dsrg_solver.set_ci_vectors(ci_vectors)
+        self.dsrg_solver.compute_gradient()
+
+        self.dsrg_cleanup()
+
+        return True
```
lcyyork, line 265: I think we either return the gradient matrix or remove the return.

forte/proc/dsrg.py
```diff
@@ -245,6 +245,26 @@ def compute_energy(self):
 
         return e_dsrg if len(self.energies) == 0 else e_relax
 
+    def compute_gradient(self, coupling_coefficients, ci_vectors):
+        """ Compute DSRG-MRPT2 analytic gradients. """
+        # Notes (Shuhe Wang):
+
+        e_relax = 0.0  # initialize this to avoid PyCharm warning
+
+        # Perform the initial un-relaxed DSRG
```
lcyyork, line 254: A safeguard of this function: test the reference relaxation. If the energy is not relaxed, proceed. If not, raise an error.

JonathonMisiewicz:
Please let me know when you're done with York and Francesco's comments: I'll give my pass then.

forte/mrdsrg-spin-integrated/master_mrdsrg.cc
```diff
@@ -177,14 +177,14 @@ void MASTER_DSRG::set_ambit_MOSpace() {
     bvirt_label_ = "V";
 
     // add Ambit index labels
-    BTF_->add_mo_space(acore_label_, "m,n,c0,c1,c2,c3,c4,c5,c6,c7,c8,c9", core_mos_, AlphaSpin);
-    BTF_->add_mo_space(bcore_label_, "M,N,C0,C1,C2,C3,C4,C5,C6,C7,C8,C9", core_mos_, BetaSpin);
-    BTF_->add_mo_space(aactv_label_, "u,v,w,x,y,z,1,2,3,a0,a1,a2,a3,a4,a5,a6,a7,a8,a9", actv_mos_,
+    BTF_->add_mo_space(acore_label_, "m,n,m1,n1,m2,n2,c0,c1,c2,c3,c4,c5,c6,c7,c8,c9", core_mos_, AlphaSpin);
```
lcyyork, line 180: I prefer to keep the original indices. I think it is relatively easy to replace `m1` to `c0`, `n1` to `c1`, etc in the code.

forte/mrdsrg-spin-integrated/dsrg_mrpt2.h
```diff
@@ -175,6 +178,190 @@ class DSRG_MRPT2 : public MASTER_DSRG {
     /// Unitary matrix to block diagonal Fock
     ambit::BlockedTensor U_;
 
+
+    // NOTICE These are essential variables and functions for computing DSRG-MRPT2 gradient.
+    // Some variables may be redundant thus need further elimination
```
lcyyork, line 182: Please do so. And also add comments.

forte/mrdsrg-spin-integrated/dsrg_mrpt2_gradient.cc
```diff
@@ -0,0 +1,3996 @@
...
+    // Tau * [1 - e^(-s * Delta^2)]
+    Tau1["ijab"] = Tau2["ijab"] * Eeps2_m1["ijab"];
+    Tau1["IJAB"] = Tau2["IJAB"] * Eeps2_m1["IJAB"];
+    Tau1["iJaB"] = Tau2["iJaB"] * Eeps2_m1["iJaB"];
+
+    outfile->Printf("Done");
+}
+
+void DSRG_MRPT2::set_sigma() {
+    outfile->Printf("\n    Initializing Sigma .............................. ");
```
lcyyork, line 907: Please be more specific on the printing, like `Initializing multipliers for ...`.

forte/mrdsrg-spin-integrated/dsrg_mrpt2.h
```diff
@@ -175,6 +178,190 @@ class DSRG_MRPT2 : public MASTER_DSRG {
...
+   
+    std::vector<size_t> core_all_;
+    /// List of active MOs (Absolute)
+    std::vector<size_t> actv_all_;
+    std::vector<size_t> virt_all_;
+
+    /// List of relative core MOs
+    std::vector<std::pair<unsigned long, unsigned long>,
+                std::allocator<std::pair<unsigned long, unsigned long>>>
+        core_mos_relative;
```
lcyyork, line 257: Please follow the naming convention where class variables are ended with `_`.

forte/base_classes/coupling_coefficients.h
lcyyork: Yes, it is a container for coupling coefficients, similar to the RDMs class. Unfortunately we need this right now for the pilot implementation. A more efficient implementation would be directly sigma build so that this class can be eliminated.

forte/base_classes/active_space_method.h
```diff
@@ -133,6 +134,17 @@ class ActiveSpaceMethod {
...
     virtual void set_options(std::shared_ptr<ForteOptions> options) = 0;
 
+    /// Return the coupling coefficients in ambit::Tensor format
+    virtual CouplingCoefficients coupling_coefficients(int level) {
+        throw std::runtime_error("Not Implemented other than FCI_MO! level = " +
+                                 std::to_string(level));
+    }
+
+    /// Return the eigen vectors
+    virtual std::vector<ambit::Tensor> eigen_vectors() {
```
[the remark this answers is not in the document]
ustcwsh: will replace them

forte/proc/dsrg.py
ustcwsh: I agree. But yesterday when I tried to remove this line, the code failed.

forte/proc/dsrg.py
ustcwsh: probably remove return in this case

forte/pymodule.py
ustcwsh: It used to be like the way you wrote, but I just thought the first condition was redundant.

forte/mrdsrg-spin-integrated/master_mrdsrg.cc
ustcwsh: We can change that after the frozen orbital version is implemented, or you are simply introducing additional difficulty to my work at this stage. Imagine the equation indices are different from the one in the codes. Additional indices make the gradient code clear and easier to understand.

forte/mrdsrg-spin-integrated/master_mrdsrg.h
ustcwsh: FIXED

forte/proc/dsrg.py
ustcwsh: Like the T2_, need be passed from dsrg_mrpt2.cc

forte/pymodule.py
ustcwsh: Sorry York, maybe I misunderstood your previous comment, but do you still want the energy computation to exist?

forte/mrdsrg-spin-integrated/active_dsrgpt2.h
ustcwsh: You may notice that such definition appears in multiple dsrg files. Actually you suggested me to do so 1 year ago, and the fact is that the code doesn't work anymore if these codes are removed. There will be some errors like :ImportError: dlopen(/Users/shuhe.wang/src/forte/forte/forte.cpython-36m-darwin.so, 2): Symbol not found: __ZN5forte10DSRG_MRPT316compute_gradientEv
  Referenced from: /Users/shuhe.wang/src/forte/forte/forte.cpython-36m-darwin.so
  Expected in: flat namespace

forte/mrdsrg-spin-integrated/master_mrdsrg.cc
lcyyork: Well, the equations in the paper are definitely different from the implementation now. OK, let's keep the change.

forte/mrdsrg-spin-integrated/active_dsrgpt2.h
lcyyork: I am surprised because this class is not even part of `MASTER_DSRG` and why do you actually need `compute_gradient()`? Please be careful with the resolve button because the suggestion is clearly not taken.

forte/pymodule.py
lcyyork: You should probably test the following works as intended. Suppose I set `job_type casscf`, CASSCF will compute gradient and write bunch of densities. Do we need this extra computation? Of course it will be minimal time compared to the PT2 gradients.

forte/pymodule.py
lcyyork: Yes, because it is confusing that `compute_gradient` actually does `compute_energy`.

forte/proc/dsrg.py
lcyyork: This should be done outside but not here.

forte/base_classes/coupling_coefficients.h
ustcwsh: @fevangelista I have changed the class name to CICouplingCoefficients

forte/proc/dsrg.py
ustcwsh: Well, when the gradient('forte') is called, the dertype is automatically set to "first" in pymodule.py. psi4.core.set_global_option('DERTYPE', 'FIRST'),
so I think we don't need to add that condition.

lcyyork: Wait, I am talking about reference relaxation here.

I see your solution works now because inside this function there is `compute_energy` but I am trying to argue `compute_energy` should not live inside `compute_gradient`.

ustcwsh: solved

forte/proc/dsrg.py
ustcwsh: solved

forte/pymodule.py
ustcwsh: In fact, the CASSCF gradient contribution is considered in the dsrggradient code by default.

ustcwsh: and thus the previous casscf gradient code is not called

forte/mrdsrg-spin-integrated/dsrg_mrpt2.h
ustcwsh: fixed

forte/sci/fci_mo.cc
```diff
@@ -2702,4 +2705,1203 @@ void FCI_MO::print_occupation_strings_perirrep(
...
+    // Get the references to the coupling lists
+    std::vector<std::vector<std::pair<size_t, short>>>& a_list = op->a_list_;
+
+    // diagonal <I| u^+ u |I>
+    for (size_t I = 0; I < ndets; ++I) {
+        std::vector<int> aocc = dets[I].get_alfa_occ(nactv_);
+        size_t idx_I = I * na3 + I * na2;
+        for (size_t u = 0, max_u = aocc.size(); u < max_u; ++u) {
+            int nu = aocc[u];
+            out.data()[idx_I + nu * nactv_ + nu] = 1.0;
```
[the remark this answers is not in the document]
ustcwsh: looks neat

forte/mrdsrg-spin-integrated/dsrg_mrpt2.h
ustcwsh: I want keep the variables unchanged for now until the frozen orbital approximation is done.

forte/mrdsrg-spin-integrated/dsrg_mrpt2_gradient.cc
ustcwsh: fixed

forte/mrdsrg-spin-integrated/dsrg_mrpt2_gradient.cc
```diff
@@ -0,0 +1,3996 @@
+/**
```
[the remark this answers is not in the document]
ustcwsh: will fix soon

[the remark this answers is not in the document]
ustcwsh: fixed

forte/api/forte_python_module.cc
```diff
@@ -106,7 +106,9 @@ void export_ActiveSpaceSolver(py::module& m) {
              "Set the active space integrals manually")
         .def("compute_fosc_same_orbs", &ActiveSpaceSolver::compute_fosc_same_orbs)
         .def("state_filename_map", &ActiveSpaceSolver::state_filename_map)
-        .def("dump_wave_function", &ActiveSpaceSolver::dump_wave_function);
+        .def("dump_wave_function", &ActiveSpaceSolver::dump_wave_function)
```
JonathonMisiewicz, line 109: Add docstrings for all the newly bound methods.

forte/base_classes/active_space_solver.cc
```diff
@@ -826,4 +827,13 @@ compute_average_state_energy(const std::map<StateInfo, std::vector<double>>& sta
     return average_energy;
 }
 
+CICouplingCoefficients ActiveSpaceSolver::coupling_coefficients(const StateInfo& state,
+                                                                int max_level) {
+    return state_method_map_[state]->coupling_coefficients(max_level);
```
JonathonMisiewicz, line 832: I'd prefer the use of `at` in place of square brackets. Sadly, we cannot guarantee that `state` will actually be in `state_method_map_`.

forte/base_classes/coupling_coefficients.h
```diff
@@ -0,0 +1,61 @@
+#ifndef _coupling_coefficients_h_
+#define _coupling_coefficients_h_
+
+#include <ambit/tensor.h>
+
+namespace forte {
+
+class CICouplingCoefficients {
```
JonathonMisiewicz, line 8: Please document what coupling coefficients _are_.

forte/base_classes/active_space_solver.h
```diff
@@ -130,6 +131,12 @@ class ActiveSpaceSolver {
     /// Set if read wave function from file as initial guess
     void set_read_initial_guess(bool read_guess) { read_initial_guess_ = read_guess; }
 
+    /// Return the coupling coefficients for a given state
+    CICouplingCoefficients coupling_coefficients(const StateInfo& state, int max_level = 2);
```
JonathonMisiewicz, line 135: I believe both these functions can be marked `const`.

JonathonMisiewicz (commented):
I made it part way through the code, but I need clarification on what coupling coefficients are. This is _not_ a phrase you use in your paper. More comments to come.

forte/api/forte_python_module.cc
lcyyork: Added.

forte/base_classes/active_space_solver.cc
lcyyork: Changed to `at` and marked as `const`.

forte/base_classes/coupling_coefficients.h
lcyyork: Do you want a real documentation or simply comment the code?

forte/base_classes/active_space_solver.h
lcyyork: To do this, I have to use `at` as in the previous comment.

forte/sci/fci_mo.cc
```diff
@@ -2703,4 +2706,777 @@ void FCI_MO::print_occupation_strings_perirrep(
...
+        int naocc = aocc.size();
+        int nbocc = bocc.size();
+
+        // aaa
+        for (int p = 0; p < naocc; ++p) {
+            int pp = aocc[p];
+            for (int q = p + 1; q < naocc; ++q) {
+                int qq = aocc[q];
+                for (int r = q + 1; r < naocc; ++r) {
+                    int rr = aocc[r];
```
JonathonMisiewicz, line 2952: Any ideas why we're not getting code coverage here?

forte/sci/fci_mo.h
```diff
@@ -264,6 +267,19 @@ class FCI_MO : public ActiveSpaceMethod {
     /// Return the parsed state-averaged info
     std::vector<std::tuple<int, int, int, std::vector<double>>> sa_info() { return sa_info_; }
 
+    /// Return 1 coupling coefficients
+    std::vector<ambit::Tensor> coupling_coeffcients_1();
```
JonathonMisiewicz, line 271: This and the below two functions are `const`, no?

forte/mrdsrg-spin-integrated/dsrg_mrpt2.h
JonathonMisiewicz: All variables should end with `_` to signify that they are class variables and not local function variables.

forte/mrdsrg-spin-integrated/dsrg_mrpt2_grad/dsrg_mrpt2_deriv_tensors.cc
```diff
@@ -0,0 +1,515 @@
...
+                            "normal Dimention-reduced Electron Repulsion Integral alpha", {"gg"});
+    V_N_Beta = BTF_->build(CoreTensor, "normal Dimention-reduced Electron Repulsion Integral beta",
+                           {"gg"});
+    V_R_Beta = BTF_->build(
+        CoreTensor, "index-reversed Dimention-reduced Electron Repulsion Integral beta", {"GG"});
+    V_all_Beta = BTF_->build(
+        CoreTensor, "normal Dimention-reduced Electron Repulsion Integral all beta", {"GG"});
+
+    // Summation of V["pmqm"] over index "m" or V["mpmq"] over index "m"
+    V_N_Alpha["pq"] = V["pmqn"] * I["mn"];
```
JonathonMisiewicz, line 360: "Normal" and "Reversed" is not standard terminology. Do you mean Coulomb and Exchange?

forte/mrdsrg-spin-integrated/dsrg_mrpt2_grad/dsrg_mrpt2_deriv_tensors.cc
```diff
@@ -0,0 +1,515 @@
...
+namespace forte {
+
+/**
+ * Initializing CI integrals for the z-vector equation.
+ */
+void DSRG_MRPT2::set_ci_ints() {
+    cc = coupling_coefficients_;
+    ci = ci_vectors_[0];
+
+    cc1a_n = ambit::Tensor::build(ambit::CoreTensor, "cc1a_n", {ndets, na, na});
```
JonathonMisiewicz, line 49: The meanings of n and r still aren't clear to me.

forte/mrdsrg-spin-integrated/dsrg_mrpt2.h
```diff
@@ -175,6 +178,187 @@ class DSRG_MRPT2 : public MASTER_DSRG {
...
+    ambit::BlockedTensor W;
+
+    // core Hamiltonian
+    ambit::BlockedTensor H;
+    // two-electron integrals
+    ambit::BlockedTensor V;
+    // Fock matrix
+    ambit::BlockedTensor F;
+
+    ambit::BlockedTensor Eeps1;
```
JonathonMisiewicz, line 286: Every single intermediate needs to be defined (mathematically) in the header file. If I'm reading your code and trying to figure out what `Eeps1` is, the header is the place I'd look.

forte/mrdsrg-spin-integrated/dsrg_mrpt2_grad/dsrg_mrpt2_deriv_tensors.cc
```diff
@@ -0,0 +1,515 @@
...
+    cc3aaa_n("Kuvwxyz") = cc.cc3aaa()("KJuvwxyz") * ci("J");
+    cc3aaa_r("Kuvwxyz") = cc.cc3aaa()("JKuvwxyz") * ci("J");
+    cc3bbb_n("KUVWXYZ") = cc.cc3bbb()("KJUVWXYZ") * ci("J");
+    cc3bbb_r("KUVWXYZ") = cc.cc3bbb()("JKUVWXYZ") * ci("J");
+    cc3aab_n("KuvWxyZ") = cc.cc3aab()("KJuvWxyZ") * ci("J");
+    cc3aab_r("KuvWxyZ") = cc.cc3aab()("JKuvWxyZ") * ci("J");
+    cc3abb_n("KuVWxYZ") = cc.cc3abb()("KJuVWxYZ") * ci("J");
+    cc3abb_r("KuVWxYZ") = cc.cc3abb()("JKuVWxYZ") * ci("J");
+
+    dlamb_aa =
```
JonathonMisiewicz, line 91: If ever you are referring to a specific equation from your paper (and I think you are here?), please give the exact number for reference.

JonathonMisiewicz (commented):
Next batch of suggestions. Any time an equation from the paper is used, I'd like the eq. number in the code for reference. That's going to be a major point. I don't anticipate major changes after that, but I'll wait to finish my review until after I have reference equation numbers.

forte/base_classes/coupling_coefficients.h
lcyyork: The index convention is actually documented in the function that computes these CCs. I can move it here but I am not sure if this container is responsible for that.

Coupling coefficients are simply the elements without summing over integral indices when computing CI matrix.

JonathonMisiewicz: Oh! Simply saying "the caller defines the indexing conventions" would be good.

forte/base_classes/coupling_coefficients.h
```diff
@@ -0,0 +1,74 @@
...
+ * We have |Psi_0> = sum_{i} C_i * |phi_i>, where C is the CI coefficient.
+ * cc1_: <phi| p^ q| Psi_0>
+ * cc2_: <phi| p^ q^ r s| Psi_0>
+ * cc3_: <phi| o^ p^ q^ r s t| Psi_0>
+ * a and b are for alpha and beta, respectively.
+ *
+ * Coupling coefficients are paired with the CI-related Lagrange multiplier
+ * in the DSRG gradient code when solving the z-vector equation.
+ */
+class CICouplingCoefficients {
```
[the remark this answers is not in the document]
lcyyork: Good question. I think we can implemented as a map but eventually we will contract the indices p,q,r,s with some kind of integrals, which are ambit::Tensors. But like I said, this class is only a container and should be deleted after we have a iterative solver for the CI part of the DSRG-MRPT2 gradients.

forte/base_classes/active_space_solver.h
```diff
@@ -130,6 +131,12 @@ class ActiveSpaceSolver {
     /// Set if read wave function from file as initial guess
     void set_read_initial_guess(bool read_guess) { read_initial_guess_ = read_guess; }
 
+    /// Return the coupling coefficients for a given state
+    CICouplingCoefficients coupling_coefficients(const StateInfo& state, int max_level = 2) const;
```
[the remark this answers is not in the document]
lcyyork: Because `ActiveSpaceSolver` has the information of determinant space. For example, CASCI and ACI hold different determinants.

The coupling coefficients can be formed between any two determinants. But we just care about the active space part, i.e., a CAS, RAS, ACI, etc. So, if we ever have a CISD class, the CISD class can also output a `CICouplingCoefficients`.

forte/pymodule.py
```diff
@@ -606,9 +618,11 @@ def gradient_forte(name, **kwargs):
 
     # Run a method
     job_type = options.get_str('JOB_TYPE')
+    correlation_solver = options.get_str('CORRELATION_SOLVER')
 
-    if job_type not in {"CASSCF", "MCSCF_TWO_STEP"}:
-        raise Exception('Analytic energy gradients are only implemented for job_types CASSCF and MCSCF_TWO_STEP.')
+    if job_type not in {"CASSCF", "MCSCF_TWO_STEP"} and correlation_solver != 'DSRG-MRPT2':
```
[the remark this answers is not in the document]
lcyyork: I think the new API will solve this problem.

forte/base_classes/active_space_solver.h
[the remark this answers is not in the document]
JonathonMisiewicz: I wrote that when I didn't realize that there was a sum over determinants. I meant to delete this, apologies.

forte/sci/fci_mo.cc
lcyyork: Not sure why, but the test `dsrg-mrpt2-gradient-1` uses CAS(4,4) and should have tested this.

forte/sci/fci_mo.h
lcyyork: Yes, I have added `const` for all.

forte/mrdsrg-spin-integrated/dsrg_mrpt2.h
```diff
@@ -175,6 +178,187 @@ class DSRG_MRPT2 : public MASTER_DSRG {
...
+    // NOTICE These are essential variables and functions for computing DSRG-MRPT2 gradient.
+    // Some variables may be redundant thus need further elimination
+    /// Set ambit tensor labels
+    void set_ambit_space();
+    /// Set tensors
+    void set_tensor();
+    /// Set density
+    void set_density();
+    /// Set Fock matrix
+    void set_fock();
```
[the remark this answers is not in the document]
ustcwsh: This function is marked as deprecated and will be removed once the frozen-orbital approximation is done, so we can leave it here for now.

forte/mrdsrg-spin-integrated/dsrg_mrpt2_grad/dsrg_mrpt2_deriv_tensors.cc
```diff
@@ -0,0 +1,515 @@
...
+        CoreTensor, "normal Dimention-reduced Electron Repulsion Integral all beta", {"GG"});
+
+    // Summation of V["pmqm"] over index "m" or V["mpmq"] over index "m"
+    V_N_Alpha["pq"] = V["pmqn"] * I["mn"];
+    // Summation of V["pMqM"] over index "M"
+    V_N_Beta["pq"] = V["pMqN"] * I["MN"];
+    // Summation of V["mPmQ"] over index "m"
+    V_R_Beta["PQ"] = V["mPnQ"] * I["mn"];
+    // Summation of V["PMQM"] over index "M"
+    V_all_Beta["PQ"] = V["PMQN"] * I["MN"];
```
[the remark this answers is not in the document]
ustcwsh: This treatment is based on the spin cases definition in forte. We only have "aaaa", "bbbb" and "abab" stored. the "Beta" means if any index is of the beta case. For all cases, we have "aaaa", "bbbb", "abab" and "baba", so we have 1 Alpha and 3 Beta.

forte/mrdsrg-spin-integrated/dsrg_mrpt2_grad/dsrg_mrpt2_deriv_tensors.cc
ustcwsh: n is for "normal" and r is for "reversed". It can be identified when checked the definition, for example     
cc1a_n("Kzw") = cc.cc1a()("KJzw") * ci("J");
cc1a_r("Kzw") = cc.cc1a()("JKzw") * ci("J");
BTW, as York mentioned, this may be deprecated in the future as well when we figure out more efficient contraction routines.

forte/mrdsrg-spin-integrated/dsrg_mrpt2.h
ustcwsh: added in .h files

forte/mrdsrg-spin-integrated/dsrg_mrpt2_grad/dsrg_mrpt2_gradient.cc
```diff
@@ -0,0 +1,137 @@
...
+#include "ci_rdm/ci_rdms.h"
+#include "boost/format.hpp"
+#include "sci/fci_mo.h"
+#include "fci/fci_solver.h"
+#include "helpers/printing.h"
+#include "../dsrg_mrpt2.h"
+
+#include "psi4/libmints/factory.h"
+#include "psi4/libiwl/iwl.hpp"
+#include "psi4/libpsio/psio.hpp"
```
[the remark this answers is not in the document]
ustcwsh: removed.

forte/mrdsrg-spin-integrated/dsrg_mrpt2_grad/dsrg_mrpt2_deriv_tensors.cc
ustcwsh: Good point. I leave comments like "derivatives of Lambda3_ w.r.t. C_K alpha-alpha-alpha", which clearly identifies what those equations are for. I think information like that is sufficient since these equation terms won't get changed in the future maintenance, as the definition of density or cumulant tensors is not changed.

forte/proc/dsrg.py
```diff
@@ -245,6 +250,15 @@ def compute_energy(self):
 
         return e_dsrg if len(self.energies) == 0 else e_relax
 
+    def compute_gradient(self, coupling_coefficients, ci_vectors):
+        """ Compute DSRG-MRPT2 analytic gradients. """
```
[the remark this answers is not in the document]
ustcwsh: added

forte/mrdsrg-spin-integrated/dsrg_mrpt2_grad/dsrg_mrpt2_deriv_tensors.cc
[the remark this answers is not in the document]
JonathonMisiewicz: Unless this naming style is the standard throughout _all of Forte_, change it to something less convoluted. `V_sumA_Beta`, `V_sumB_Beta`, `V_sumA_Alpha`, `V_sumB_Alpha`, for example, is much cleaner. (Do you even need to separate out `V_sumB_Beta` and `V_sumA_Beta`? I haven't looked at the equations, but I would be astounded if you do.)

forte/mrdsrg-spin-integrated/dsrg_mrpt2.h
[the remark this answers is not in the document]
JonathonMisiewicz: Shuhe, this is a single search and replace `sed` command. You could have made that change in the time it took you to type that comment.

forte/mrdsrg-spin-integrated/dsrg_mrpt2_grad/dsrg_mrpt2_deriv_tensors.cc
JonathonMisiewicz: I find "n" and "r" confusing, especially because there is no apparent relation between this and the contractions of the 1RDM with the G integrals, where you also use "n" and "r". Can we label these `bra` and `ket`?

Shouldn't the coupling coefficients be symmetric? If so, `cc1a_n == cc1a_r`.

forte/mrdsrg-spin-integrated/dsrg_mrpt2_grad/dsrg_mrpt2_gradient.cc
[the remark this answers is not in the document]
JonathonMisiewicz: Removing _duplicate_ headers is a start, but you need to remove the _unused_ headers.

forte/mrdsrg-spin-integrated/dsrg_mrpt2_grad/dsrg_mrpt2_deriv_tensors.cc
JonathonMisiewicz: And for this section, that's fine (assuming this is the reference cumulant?), but what about all the other equations?

JonathonMisiewicz:
This issue keeps coming, so it's probably worth making this an explicit point of discussion:
How important is it that this code be maintainable, given that large parts of it may be replaced? It would help if I had a rough timeline on this.

forte/mrdsrg-spin-integrated/dsrg_mrpt2_grad/dsrg_mrpt2_deriv_tensors.cc
ustcwsh: "Smart" idea but it is wrong. When you set cc1a_n == cc1a_r you get the wrong gradient, that's one problem spotted and solved years ago.

forte/mrdsrg-spin-integrated/dsrg_mrpt2.h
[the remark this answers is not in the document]
ustcwsh: Yeah thanks for teaching me that stream editor command "sed" which you think I haven't learnt before.

ustcwsh:
> Next batch of suggestions. Any time an equation from the paper is used, I'd like the eq. number in the code for reference. That's going to be a major point. I don't anticipate major changes after that, but I'll wait to finish my review until after I have reference equation numbers.

I would be happy to, if its the CASSCF gradient or other easy stuffs. Two points: 1, We basically have over 300 equations, it's not a easy task (talking about it is absolutely much easier); 2, Different DSRG-MRPT2 energy terms are properly separated with comments and the whole gradient calculation procedure is also split into multiple individual functions. You can easily locate the term or equations you are looking at; 3, for most of time, you don't have to customize or change anything in those equations.

ustcwsh:
> This issue keeps coming up, so it's probably worth making this an explicit point of discussion:
> How important is it that this code be maintainable, given that large parts of it may be replaced? It would help if I had a rough timeline on _when_ this replacement would occur.

The majority of the code is about tensor contraction. Since the DSRG-MRPT2 energy expression itself won't change and the way ambit performs tensor contractions hasn't been changed for years, I don't think "large parts of it may be replaced" makes any sense to me.

forte/mrdsrg-spin-integrated/dsrg_mrpt2_grad/dsrg_mrpt2_deriv_tensors.cc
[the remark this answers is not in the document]
ustcwsh: changed.

forte/mrdsrg-spin-integrated/dsrg_mrpt2.h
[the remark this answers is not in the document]
ustcwsh: fixed

forte/mrdsrg-spin-integrated/dsrg_mrpt2_grad/dsrg_mrpt2_deriv_tensors.cc
ustcwsh: changed.

forte/mrdsrg-spin-integrated/dsrg_mrpt2_grad/dsrg_mrpt2_gradient.cc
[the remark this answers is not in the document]
ustcwsh: all files are cleaned now

tests/methods/dsrg-mrpt2-gradient-1/input.dat
```diff
@@ -0,0 +1,50 @@
+# DSRG-MRPT2 gradient on 4 H atoms with c1 symmetry
+import forte
+
+ref_grad = psi4.Matrix.from_list([
+      [-0.521586919763,    -1.188829187079,    -0.914234910278],
```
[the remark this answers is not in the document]
JonathonMisiewicz: I still recommend this change.

forte/mrdsrg-spin-integrated/dsrg_mrpt2.h
```diff
@@ -175,6 +178,294 @@ class DSRG_MRPT2 : public MASTER_DSRG {
...
+    std::vector<std::pair<unsigned long, unsigned long>,
+                std::allocator<std::pair<unsigned long, unsigned long>>>
+        actv_mos_relative;
+    /// List of relative virtual MOs
+    std::vector<std::pair<unsigned long, unsigned long>,
+                std::allocator<std::pair<unsigned long, unsigned long>>>
+        virt_mos_relative;
+
+    /// Dimension of different irreps
+    psi::Dimension irrep_vec;
```
JonathonMisiewicz, line 374: There are a _lot_ of class variables that still need to be renamed to end in an `_`. I know it's tedious, but it's an easy way to check if a variable is a class variable without checking the entire header file.

forte/mrdsrg-spin-integrated/dsrg_mrpt2.h
JonathonMisiewicz: Shuhe, could you respond to this?

forte/mrdsrg-spin-integrated/dsrg_mrpt2_grad/dsrg_mrpt2_deriv_write_rdms.cc
```diff
@@ -0,0 +1,553 @@
...
+    outfile->Printf("Done");
+}
+
+void DSRG_MRPT2::tpdm_backtransform() {
+    // Backtransform the TPDM
+
+    std::vector<std::shared_ptr<psi::MOSpace>> spaces;
+    spaces.push_back(psi::MOSpace::all);
+    std::shared_ptr<TPDMBackTransform> transform =
+        std::shared_ptr<TPDMBackTransform>(new TPDMBackTransform(
```
JonathonMisiewicz, line 541: It's better practice to use `make_shared` here. It saves a little bit of computer time, rather than make a raw pointer with new and then cast to `shared_ptr`.

JonathonMisiewicz (commented):
I'm mostly happy with this! There's some polishing to do to take this from a pilot implementation to a production-level implementation, but that can come later. For posterity:

1. Right now, we have _many_ extra class variables with DSRG-MRPT2. Shuhe said we should wait to prune those until after frozen core and frozen virtual. I can accept that, but we will need to revisit the issue then. Even then, I'm wondering if we should have a separate DSRG-MRPT2 gradient computer class, so those variables never get touched until it's time for the gradient.
2. When the paper is more stable, we should go back and add exact equation numbers in comments.
3. We will definitely need to revisit CouplingCoefficents to implement something more efficient. Even the current class has what I believe are some redundancies. (Bra and Ket vectors should be _transposes_ of each other.)

forte/base_classes/coupling_coefficients.h
```diff
@@ -0,0 +1,79 @@
...
+ *
+ * NOTE: 1. Any production-level code should NOT use this class!
+ *       2. Users are responsible for the indexing conventions!
+ */
+class CICouplingCoefficients {
+  public:
+    CICouplingCoefficients();
+
+    /// @brief Construct using 1 and 2 coupling coefficients
+    CICouplingCoefficients(ambit::Tensor cc1a, ambit::Tensor cc1b, ambit::Tensor cc2aa,
```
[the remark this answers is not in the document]
lcyyork: I think these will be copied by pointers.

forte/mrdsrg-helper/dsrg_source.h
```diff
@@ -47,9 +47,16 @@ class DSRG_SOURCE {
 
     /// Bare effect of source operator
     virtual double compute_renormalized(const double& D) = 0;
-    /// Renormalize denominator
+
+    /// Regularize denominator
     virtual double compute_renormalized_denominator(const double& D) = 0;
 
+    /// Derivative of regularized denominator w.r.t. denominator (regularized part)
+    /// TODO: probably not general and only useful for STD and LABS
```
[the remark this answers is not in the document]
lcyyork: These are different source operators.

The function I would like to have is `compute_regularized_denominator_deriv`, which takes the derivative of the regularizer with respect to the energy denominator. For the standard one (1-e^{-s D^2}) / D, there are two parts in the derivative where one puts D^2 in the denominator while the other one does not care about small D values. This function `compute_regularized_denominator_derivR` implements that contains 1/D^2.

forte/mrdsrg-spin-integrated/dsrg_mrpt2_grad/dsrg_mrpt2_deriv_solve_zvector.cc
```diff
@@ -970,6 +973,71 @@ void DSRG_MRPT2::z_vector_contraction(std::vector<double> & qk_vec, std::vector<
...
 
+    /// Call the generalized sigma function to complete the contraction
+    for (const auto& pair: as_solver_->state_space_size_map()) {
+        const auto& state = pair.first;
+        std::map<std::string, double> block_factor;
+        block_factor["aa"] = 1.0;
+        block_factor["AA"] = 1.0;
+
+
+        auto antisym_1 = BTF_->build(CoreTensor, "antisymmetrized 1-body intermediate tensor", spin_cases({"aa"}));
```
lcyyork, line 987: I think there is a misunderstanding of my slack message. There is nothing to be antisymmetrized for 1-body integrals.

forte/mrdsrg-spin-integrated/dsrg_mrpt2_grad/dsrg_mrpt2_deriv_solve_zvector.cc
```diff
@@ -970,6 +973,71 @@ void DSRG_MRPT2::z_vector_contraction(std::vector<double> & qk_vec, std::vector<
...
+            temp_1["UV"] += 2 * V_sumA_Beta["VE"] * qk["EU"];
+            // active-active
+            temp_1["xy"] += V["uyvx"] * qk["uv"];
+            temp_1["XY"] += V["uYvX"] * qk["uv"];
+            // beta
+            temp_1["XY"] += V["UYVX"] * qk["UV"];
+            temp_1["xy"] += V["yUxV"] * qk["UV"];
+
+ 
+            /// Antisymmetrization
```
lcyyork, line 1024: This is not antisymmetrization. You are making it Hermitian here.

PR #255, opened 2021-06-18
Simplify the orthonormality checker.

JonathonMisiewicz:
## Description
This PR simplifies the orthonormality checker. Because we are constructing one less MintsHelper object, the output file is a bit less cluttered with basisset construction.

## Checklist
- [x] `run_forte_tests.py` passes
- [x] Ready to go!


fevangelista (approved):
Looks like a good way to simplify the code and it's probably good to separate building S from computing S_MO.

PR #256, opened 2021-06-19
New python interface: Active space solvers

fevangelista:
## Description
This PR introduces the following new functionality:

- **Added new functionality and refactored the Solver class to make it work as a node in a graph**: Initialization of some objects is now routed to the base class `Solver`.
- **Improved contract between `Solver` objects**: Each `Solver` object now holds a list of the input nodes, and of the features that it provides and needs. These can be used to test if a computational graph is correct at the time of creating the graph. The function `computational_graph` has been added to print out a computational graph (only works with simple graphs, no branching).
- **Code to setup and run computations using Forte's active space solvers (FCI, ACI, ...)**. This is a simple matter of hiding python code that sets up an active space solver.
The current working design implements a class that can handle all active space solvers
```python
class ActiveSpaceSolver(Solver):
    ...
```
- **A class to perform spin analysis**. This is added to provide spin analysis. In the new design it can be used with any node that can provide RDMs.
- **Logging of forte**. Disabled by default, logging can be activated and used to print out information about the computation flow.
- **Handling of density fitting**. In the new driver the integral type is set at the level of the model, so this should prevent issues where a user forgets to turn on DF at some state of a computation.
- **Updates to the options class**: Added copy constructor and changed `set_dict` to make sure options object do a deep copy of the options (old code does a shallow copy, preventing making a different copy of an option object). This change is helpful to inject options at different stages of the computation (without having to stash options and retrieve them).
 
The following is an example of a FCI computation.
```python
from forte.solvers import HF, FCI, solver_factory

xyz = """
H 0.0 0.0 0.0
H 0.0 0.0 1.0
"""

# create the input node of the graph and populate it with a molecular model
input = solver_factory(molecule=xyz, basis='cc-pVDZ', int_type='DF')

# specify the electronic state
state = input.state(charge=0, multiplicity=1, sym='ag')

# create a HF object
hf = HF(input, state=state)

# create a FCI object that grabs the MOs from the HF object (hf)
fci = ActiveSpaceSolver(hf, type='FCI', states=state, active=[1, 0, 0, 0, 0, 1, 0, 0])

# run the computation
fci.run()
```
This code is equivalent to instantiating the following computational graph:
image

## User Notes
- [x] Translated psi4-based input for several test to new api. These examples were added to `/tests/pytest-methods`
- [x] Added new tests for functions in the options class

## Checklist
- [x] Documented source code
- [x] Ready to go!


JonathonMisiewicz:
Just skimming the commits, it looks like this PR has expanded from the original scope. When the time comes for reviews, please expand the title post to discuss:

- [ ] New logging functionality
- [ ] Changes in the options system

forte/core.py
```diff
@@ -66,3 +69,40 @@ def __new__(cls):
 
     def __del__(cls):
         forte.cleanup()
+
+
+def start_logging():
+    """This function starts logging"""
```
JonathonMisiewicz, line 75: The docstring can be more specific: sets log output to a file, and sets the log level to all information. (Default is warnings, errors, and critical errors, which we signal by terminating Forte.) `flog` below may make this distinction important.

forte/core.py
```diff
@@ -66,3 +69,40 @@ def __new__(cls):
...
+    Log the message ``msg`` with logging level ``level``.
+
+    This function calls the logging module and puts some spaces before
+    the message that we want to log.
+    """
+    global logging_depth
+    if level == 'info':
+        spaces = ' ' * max((logging_depth - 1), 0) * 2
+        logging.info(f"{spaces}{msg}")
+
```
JonathonMisiewicz, line 91: Raise an error rather than fail silently, if no logging level is set.

forte/solvers/mcscf.py
```diff
@@ -0,0 +1,85 @@
+# import psi4
```
JonathonMisiewicz, line 1: Why is this file here but commented out?

forte/api/forte_python_module.cc
```diff
@@ -162,9 +162,18 @@ PYBIND11_MODULE(forte, m) {
     m.def("make_fragment_projector", &make_fragment_projector,
           "Make a fragment(embedding) projector");
     m.def("make_embedding", &make_embedding, "Apply fragment projector to embed");
-    m.def("make_ints_from_psi4", &make_forte_integrals_from_psi4,
-          "Make Forte integral object from psi4");
-    m.def("make_custom_ints", &make_custom_forte_integrals, "Make a custom integral object");
+    m.def("make_custom_ints", &make_custom_forte_integrals,
+          "Make a custom Forte integral object from arrays");
+    m.def("make_ints_from_psi4",
```
JonathonMisiewicz, line 167: Why do we have two separate functions for this?

If we want to retire implicitly grabbing `SCF_TYPE` from options, let's add a deprecation warning. If we don't want to retire it, why not combine these two functions into a single function with an optional `std::string` argument?

forte/api/state_info_api.cc
```diff
@@ -27,6 +27,7 @@
  */
 
 #include <pybind11/pybind11.h>
+#include <pybind11/stl.h>
```
JonathonMisiewicz, line 30: I'm assuming you get compiler issues if you remove this?

forte/base_classes/forte_options.h
```diff
@@ -44,13 +44,36 @@ namespace forte {
...
+ *
+ * The options are held in python dictionary (pybind11::dict) and can
+ * be easily grabbed (via the dict() function) or set from python.
+ * When copying this object, we make sure that a deep copy is performed,
+ * otherwise a change in one object may change options in a different
+ * object from which it was obtained.
+ *
+ * Each option is held in dictionary with the following entries:
+ * - type: string the option type (e.g. "int")
+ * - group: the group to which this option belongs (may be empy "")
```
JonathonMisiewicz, line 58:
```suggestion
 * - group: the group to which this option belongs (may be empty "")
```

forte/base_classes/forte_options.cc
```diff
@@ -171,6 +176,59 @@ void ForteOptions::add_double_array(const std::string& label, const std::string&
     add(label, "float_list", py::list(), description);
 }
 
+void ForteOptions::set_options_from_dict(pybind11::dict dict) {
```
JonathonMisiewicz, line 179: I'm not sure if `const` and `&` have special types when working with Pybind, but the signatures here seem strange to me. Is there a reason we can't mark the `pybind11::dict` const? I also understand we need to copy the dict one time, but we seem to be copying it twice: once when passing the argument, and once when looping over it. Am I missing something? This could be some aspect of C++ copy semantics I'm overlooking, but I'd double-check this.

forte/base_classes/forte_options.h
```diff
@@ -44,13 +44,36 @@ namespace forte {
...
+ * - value: the value of the option
+ * - default_value: the default value given to the option
+ * - description: a description of the this option and what it controlss
```
JonathonMisiewicz, line 61:
```suggestion
 * - description: a description of the option and what it controlss
```

forte/dmrg/dmrgscf.cc
```diff
@@ -206,8 +209,8 @@ void DMRGSCF::copyCHEMPS2MXtoPSIMX(CheMPS2::DMRGSCFmatrix* source,
 }
 
 void DMRGSCF::buildQmatOCC(CheMPS2::DMRGSCFmatrix* theQmatOCC, CheMPS2::DMRGSCFindices* iHandler,
-                           psi::SharedMatrix MO_RDM, psi::SharedMatrix MO_JK, psi::SharedMatrix Cmat,
-                           std::shared_ptr<psi::JK> myJK) {
+                           psi::SharedMatrix MO_RDM, psi::SharedMatrix MO_JK,
+                           psi::SharedMatrix Cmat, std::shared_ptr<psi::JK> myJK) {
```
JonathonMisiewicz, line 213: While I understand this is beyond the scope of the PR, I think you want references throughout this file, not shared pointers.

forte/model.py
```diff
@@ -19,28 +22,42 @@ class Model(ABC):
...
+    ):
         """
         Initialize a MolecularModel object
 
         Parameters
         ----------
         molecule: Molecule
             the molecule information
+        int_type: str
+            the type of integrals (e.g., 'CONVENTIONAL', 'DF', 'CD',...)
```
JonathonMisiewicz, line 50: Standard syntax is to include the list of allowed values in place of the argument type, e.g.,
`int_type: {'CONVENTIONAL', 'DF, 'CD'}`

forte/solvers/active_space_solver.py
```diff
@@ -0,0 +1,155 @@
...
+    ):
+        """
+        Initialize a FCI object
+
+        Parameters
+        ----------
+        input: Solver
+            The solver that will provide the molecular orbitals
+        states: dict()
+            A dictionary of StateInfo -> list(float) of all the states that will be computed.
```
JonathonMisiewicz, line 43: What does this list do? I suspect something with state-averaging.

forte/solvers/active_space_solver.py
```diff
@@ -0,0 +1,155 @@
...
+        type: str
+            The type of solver (e.g. one of 'FCI', 'ACI', ...)
```
JonathonMisiewicz, line 45: Again, supply an explicit list in place of `str` for the argument type in the docstring

forte/solvers/hf.py
```diff
@@ -12,30 +11,31 @@ class HF(Solver):
...
         cbh=None
     ):
         """
-        initialize a Basis object
+        initialize a HF object
 
         Parameters
         ----------
-        restricted : bool
+        input: Solver
```
JonathonMisiewicz, line 29: Why rename this variable to input? It's vague.

forte/solvers/spin_analysis.py
```diff
@@ -0,0 +1,63 @@
...
+        return f'SpinAnalysis(options={self._options})'
+
+    def __str__(self):
+        """
+        return a string representation of this object
+        """
+        return repr(self)
+
+    def _run(self):
+        """Run a Hartree-Fock computation"""
```
JonathonMisiewicz, line 44:
```suggestion
        """Run the spin analysis"""
```

tests/methods/fci-7/input.dat
```diff
@@ -17,14 +17,16 @@ set {
   scf_type pk
   reference rohf
   e_convergence 12
+  docc [3,0,0,0]
```
JonathonMisiewicz, line 20: Why change the test case?

tests/pytest/solver/test_solver-2.py
```diff
@@ -0,0 +1,29 @@
...
+    root = solver_factory(molecule=xyz, basis='cc-pVDZ')
+
+    # specify the electronic state
+    state = root.state(charge=0, multiplicity=1, sym='ag')
+
+    # create a HF object and run
+    hf = HF(root, state=state)
+    fci = ActiveSpaceSolver(hf, state, 'FCI')
+    spin = SpinAnalysis(fci)
+    print(spin.computational_graph())
```
JonathonMisiewicz, line 25: This tests that the computation runs, but not its actual results. Same with `test_solver.py` Can you change this? These are the only tests on your new SpinAnalysis class, and I'd like some correctness tests on that.

JonathonMisiewicz (commented):
Great overall!

From a reviewer standpoint, I especially like migrating old tests to use the new API, and categorizing the various areas this PR touches.

forte/base_classes/forte_options.cc
```diff
@@ -171,6 +176,59 @@ void ForteOptions::add_double_array(const std::string& label, const std::string&
...
+            throw std::runtime_error(msg);
+        }
+    }
+}
+
+void ForteOptions::set_dict(const pybind11::dict dict) {
+    dict_ = py::module::import("copy").attr("deepcopy")(dict);
+}
+
+void ForteOptions::reset_dict() { dict_ = pybind11::dict(); }
```
lcyyork, line 230: Reset could also mean set back to default values.

forte/model.py
```diff
@@ -19,28 +22,42 @@ class Model(ABC):
     def point_group(self) -> str:
         """The model point group"""
 
+    @abstractmethod
+    def ints(self, mo_space_info):
+        """Make a ForteIntegral object"""
+
```
lcyyork, line 28:
```suggestion
        return

```

forte/register_forte_options.py
```diff
@@ -102,905 +93,805 @@ def register_avas_options(options):
...
-    options.add_int_array("GAS5", "Number of GAS5 orbitals per irrep"
-                          " (in Cotton order)")
-    options.add_int_array("GAS6", "Number of GAS6 orbitals per irrep"
-                          " (in Cotton order)")
+    options.add_int_list("FROZEN_DOCC", "Number of frozen occupied orbitals" " per irrep (in Cotton order)")
+    options.add_int_list(
+        "RESTRICTED_DOCC", "Number of restricted doubly"
+        " occupied orbitals per irrep (in Cotton order)"
+    )
+    options.add_int_list("ACTIVE", " Number of active orbitals per irrep" " (in Cotton order)")
```
lcyyork, line 207:
```suggestion
    options.add_int_list("ACTIVE", " Number of active orbitals per irrep (in Cotton order)")
```

forte/register_forte_options.py
```diff
@@ -102,905 +93,805 @@ def register_avas_options(options):
...
+    options.add_int_list("RESTRICTED_UOCC", "Number of restricted unoccupied" " orbitals per irrep (in Cotton order)")
```
lcyyork, line 208:
```suggestion
    options.add_int_list("RESTRICTED_UOCC", "Number of restricted unoccupied orbitals per irrep (in Cotton order)")
```

forte/register_forte_options.py
```diff
@@ -102,905 +93,805 @@ def register_avas_options(options):
...
+    options.add_int_list("FROZEN_UOCC", "Number of frozen unoccupied orbitals" " per irrep (in Cotton order)")
```
lcyyork, line 209:
```suggestion
    options.add_int_list("FROZEN_UOCC", "Number of frozen unoccupied orbitals per irrep (in Cotton order)")
```

forte/register_forte_options.py
```diff
@@ -102,905 +93,805 @@ def register_avas_options(options):
...
+
+    options.add_int_list("GAS1", "Number of GAS1 orbitals per irrep" " (in Cotton order)")
```
lcyyork, line 211:
```suggestion
    options.add_int_list("GAS1", "Number of GAS1 orbitals per irrep (in Cotton order)")
```

forte/register_forte_options.py
```diff
@@ -102,905 +93,805 @@ def register_avas_options(options):
...
+    options.add_int_list("GAS2", "Number of GAS2 orbitals per irrep" " (in Cotton order)")
```
lcyyork, line 212:
```suggestion
    options.add_int_list("GAS2", "Number of GAS2 orbitals per irrep (in Cotton order)")
```

forte/register_forte_options.py
```diff
@@ -102,905 +93,805 @@ def register_avas_options(options):
...
+    options.add_int_list("GAS3", "Number of GAS3 orbitals per irrep" " (in Cotton order)")
```
lcyyork, line 213:
```suggestion
    options.add_int_list("GAS3", "Number of GAS3 orbitals per irrep (in Cotton order)")
```

forte/register_forte_options.py
```diff
@@ -102,905 +93,805 @@ def register_avas_options(options):
...
+    options.add_int_list("GAS4", "Number of GAS4 orbitals per irrep" " (in Cotton order)")
+    options.add_int_list("GAS5", "Number of GAS5 orbitals per irrep" " (in Cotton order)")
+    options.add_int_list("GAS6", "Number of GAS6 orbitals per irrep" " (in Cotton order)")
```
lcyyork, line 216:
```suggestion
    options.add_int_list("FROZEN_DOCC", "Number of frozen occupied orbitals per irrep (in Cotton order)")
    options.add_int_list(
        "RESTRICTED_DOCC", "Number of restricted doubly"
        " occupied orbitals per irrep (in Cotton order)"
    )
    options.add_int_list("ACTIVE", " Number of active orbitals per irrep (in Cotton order)")
    options.add_int_list("RESTRICTED_UOCC", "Number of restricted unoccupied orbitals per irrep (in Cotton order)")
    options.add_int_list("FROZEN_UOCC", "Number of frozen unoccupied orbitals per irrep (in Cotton order)")

    options.add_int_list("GAS1", "Number of GAS1 orbitals per irrep (in Cotton order)")
    options.add_int_list("GAS2", "Number of GAS2 orbitals per irrep (in Cotton order)")
    options.add_int_list("GAS3", "Number of GAS3 orbitals per irrep (in Cotton order)")
    options.add_int_list("GAS4", "Number of GAS4 orbitals per irrep (in Cotton order)")
    options.add_int_list("GAS5", "Number of GAS5 orbitals per irrep (in Cotton order)")
    options.add_int_list("GAS6", "Number of GAS6 orbitals per irrep (in Cotton order)")
```

forte/register_forte_options.py
```diff
@@ -102,905 +93,805 @@ def register_avas_options(options):
...
-    options.add_double("PCI_SPAWNING_THRESHOLD", 0.001,
-                       "The determinant importance threshold")
+    options.add_double("PCI_SPAWNING_THRESHOLD", 0.001, "The determinant importance threshold")
 
     options.add_int(
-        "PCI_MAX_GUESS_SIZE", 10000,
-        "The maximum number of determinants used to form the "
-        "guess wave function")
+        "PCI_MAX_GUESS_SIZE", 10000, "The maximum number of determinants used to form the "
+        "guess wave function"
```
lcyyork, line 276:
```suggestion
        "PCI_MAX_GUESS_SIZE", 10000, "The maximum number of determinants used to form the guess wave function"
```

forte/register_forte_options.py
```diff
@@ -102,905 +93,805 @@ def register_avas_options(options):
...
-                     "Doing post calcualtion occupation analysis?")
+    options.add_bool("OCC_ANALYSIS", False, "Doing post calcualtion occupation analysis?")
 
-    options.add_double("OCC_LIMIT", 0.0001, "Occupation limit for considering"
-                       " if an orbital is occupied/unoccupied in the post"
-                       " calculation analysis.")
+    options.add_double(
+        "OCC_LIMIT", 0.0001, "Occupation limit for considering"
+        " if an orbital is occupied/unoccupied in the post"
+        " calculation analysis."
```
lcyyork, line 493:
```suggestion
        " if an orbital is occupied/unoccupied in the post calculation analysis."
```

forte/register_forte_options.py
```diff
@@ -102,905 +93,805 @@ def register_avas_options(options):
...
+    )
 
-    options.add_double("CORR_LIMIT", -0.01, "Correlation limit for considering"
-                       " if two orbitals are correlated in the post"
-                       " calculation analysis.")
+    options.add_double(
+        "CORR_LIMIT", -0.01, "Correlation limit for considering"
+        " if two orbitals are correlated in the post"
+        " calculation analysis."
```
lcyyork, line 499:
```suggestion
        " if two orbitals are correlated in the post calculation analysis."
```

forte/register_forte_options.py
```diff
@@ -102,905 +93,805 @@ def register_avas_options(options):
...
-                    "Iteration number to start adding error vectors"
-                    " (< 1 will not do DIIS)")
-    options.add_int("CASSCF_DIIS_FREQ", 1,
-                    "How often to do DIIS extrapolation")
-    options.add_double("CASSCF_DIIS_NORM", 1e-3,
-                       "Do DIIS when the orbital gradient norm is below this value")
+    options.add_bool("CASSCF_DO_DIIS", True, "Use DIIS in CASSCF orbital optimization")
+    options.add_int("CASSCF_DIIS_MIN_VEC", 2, "Minimum size of DIIS vectors for orbital rotations")
+    options.add_int("CASSCF_DIIS_MAX_VEC", 8, "Maximum size of DIIS vectors for orbital rotations")
+    options.add_int("CASSCF_DIIS_START", 2, "Iteration number to start adding error vectors" " (< 1 will not do DIIS)")
```
lcyyork, line 873:
```suggestion
    options.add_int("CASSCF_DIIS_START", 2, "Iteration number to start adding error vectors (< 1 will not do DIIS)")
```

forte/solvers/solver.py
```diff
@@ -1,31 +1,117 @@
...
+        flog('info', f'{type(self).__name__}: run() finished executing')
+
+        # set executed flag
+        self._executed = True
+
+        return self.data
+
+    @abstractmethod
+    def _run():
+        """The actual run function implemented by each method"""
```
lcyyork, line 102:
```suggestion
        """The actual run function implemented by each method"""
        pass
```

lcyyork (commented):
This is all good. I see the active space is now explicitly specified when forming `ActiveSpaceSolver`. I wonder if we should change that similar to `state`.

forte/api/state_info_api.cc
fevangelista: This is needed. If absent a runtime error is emitted complaining about conversion of data (IIRC).

forte/base_classes/forte_options.cc
fevangelista: I suspect that `pybind11::dict` is just a handle (a pointer) to the python object. I am going to try the standard `const ...&` way of passing, which is probably the most economical.

forte/base_classes/forte_options.cc
fevangelista: This function does not make any sense (it is equivalent to allocating a new object), so I am going to remove it.

forte/solvers/solver.py
fevangelista: I'll implement the `pass` but from what I know the minimum text needed for a function to be fine is either a docstring or `pass`, and both are equivalent in the sense that they don't perform any operation.

tests/methods/fci-7/input.dat
fevangelista: This test case is potentially ill-defined. The SCF solution was actually converging on a B1 state and the FCI was instead computing a B2 solution. This is because for this test case the B1 and B2 MOs are degenerate. The new code in the HF class picked up the discrepancy in the symmetry of the two solutions. Now, since there is only one set of p orbitals in the basis, this mixed B1/B2 computation gives the same as doing both a HF and FCI computation on the same state (say B1), however, that's just a coincidence. These extra options now precisely define the test case (without changing the result).

tests/pytest/solver/test_solver-2.py
fevangelista: Good catch. I now test the computational graph.

forte/api/forte_python_module.cc
fevangelista: I combined the two functions. Eventually we want to deprecate, but that's for another time.

forte/core.py
fevangelista: Another good point. A more detailed explanation added.

forte/core.py
fevangelista: I implemented the other logging levels and added an exception if fails.

forte/dmrg/dmrgscf.cc
fevangelista: Oh, yes. We need to leave this for another PR.

forte/model.py
fevangelista: Added.

forte/solvers/active_space_solver.py
fevangelista: Added.

forte/solvers/active_space_solver.py
fevangelista: I added a long comment on this option.

forte/solvers/hf.py
fevangelista: This is probably transitory. I am trying to stick to the terminology of nodes where you have input and output nodes. Maybe use `input_solver`? `parent_solver` is also reasonable and I could be swayed to go back to it.

forte/solvers/mcscf.py
fevangelista: Deleted.

forte/register_forte_options.py
fevangelista: Sorry for the formatting mess.

fevangelista:
@lcyyork: I am not sure I understand your last suggestion. Are you saying that it would be a good idea to create a separate object that specifies the active space that is then passed to the active space solver? If that is what you suggest I agree it would be useful and could be a good way to separate data. Would this just be a MOSpaceInfo object created separately? I found it useful when doing the rewrite to start from the interface, so I'd be interested in any code example that illustrates your idea in an actual computation.

lcyyork:
> @lcyyork: I am not sure I understand your last suggestion. Are you saying that it would be a good idea to create a separate object that specifies the active space that is then passed to the active space solver? If that is what you suggest I agree it would be useful and could be a good way to separate data. Would this just be a MOSpaceInfo object created separately? I found it useful when doing the rewrite to start from the interface, so I'd be interested in any code example that illustrates your idea in an actual computation.

@fevangelista Yes, I was thinking just passing in `mo_space_info`.

fevangelista:
@lcyyork: It turn out I cannot define a full MOSpaceInfo object before running SCF (because I need nmopi). So what I can do is let the user specify a dictionary that is then passed to the solvers. In the last commit I addressed this and I introduced a Node class to separate responsibilities. The inheritance structure is now a bit more complex:
```
Node
 |---Input
 |
 |---Solver-|
            |---HF
            |---ActiveSpaceSolver
```

fevangelista:
I have a small change to make to the manual page (https://forte-fork.readthedocs.io/en/pyf-fci/nb_api_intro.html) and then this should be ready for merge.

JonathonMisiewicz:
If you don't mind, let me know when the final commit is in. I want to give it a final pass, given the changes suggested thus far.

fevangelista:
Just pushed changes to the manual page and some documentation. So @JonathonMisiewicz this is ready for a final pass. I am sure we will have to revisit some choices, but if you spot something urgent let me know.

tests/pytest/solver/test_solver.py
```diff
@@ -0,0 +1,28 @@
...
+    H 0.0 0.0 0.0
+    H 0.0 0.0 1.0
+    """
+
+    # create a molecular model
+    input = solver_factory(molecule=xyz, basis='cc-pVDZ')
+
+    # create a HF object and run
+    with pytest.raises(AssertionError):
+        spin = SpinAnalysis(input)
```
JonathonMisiewicz, line 24: Is there any way to test that the underlying `perform_spin_analysis` is behaving as expected? I don't see coverage of this anywhere.

JonathonMisiewicz (approved):
I have one last lingering question, but I'm happy to merge whenever.

tests/pytest/solver/test_solver.py
fevangelista: I forgot to mention this: the spin analysis is tested in `tests/pytest-methods/test_aci-3.py`. This test only makes sure that we catch the user trying to assemble a nonsensical computational graph.

lcyyork (approved):

PR #259, opened 2021-06-28
Fix error message in DETCI

fevangelista:
## Description
This PR improves an error message in DETCI that is triggered if no determinants are found.


ustcwsh (approved):

ustcwsh (commented):
"root symmetry" might be more accurate, "symmetry" sounds like point group symmetry and could be misleading

PR #260, opened 2021-06-29
New python interface: MCSCF

fevangelista:
## Description
The goal of this PR is to implement a MCSCF class that can use the existing `ActiveSpaceSolver` class to optimize the orbitals.

The following is an example of a CASSCF computation enabled by this PR
```python
from forte.solvers import HF, ActiveSpaceSolver, MCSCF, input_factory

xyz = """
H 0.0 0.0 0.0
H 0.0 0.0 1.0
"""

# Create an input node
input = input_factory(molecule=xyz, basis='cc-pVDZ', int_type='DF')
state = input.state(charge=0, multiplicity=1, sym='ag')
mo_space = input.mo_space(active=[1, 0, 0, 0, 0, 1, 0, 0])

# construct the following procedure HF -> FCI -> MCSCF
hf = HF(input, state=state)
fci = ActiveSpaceSolver(hf, type='FCI', states=state, mo_space= mo_space)
mcscf = MCSCF(fci) # <- optimize the states specified in FCI
# run the computation
mcscf.run()
```

## User Notes
- [x] Added class MCSCF to run the C++ class `MCSCF_2STEP`.
- [x] Enabled DF with MCSCF
- [x] Enabled multiple active space solvers with MCSCF 

## Checklist
- [x] Translated several tests from psithon input and added some new
- [x] Documented source code
- [ ] All tests pass 
- [x] Ready to go!


JonathonMisiewicz (commented):
A very preliminary review, because the "big picture" here confuses me. Why are we requiring an active space computation to be done _between_ the HF and the MCSCF?

forte/api/forte_python_module.cc
```diff
@@ -325,6 +324,9 @@ PYBIND11_MODULE(forte, m) {
     py::class_<DressedQuantity>(m, "DressedQuantity")
         .def("contract_with_rdms", &DressedQuantity::contract_with_rdms, "reference"_a,
              "Contract densities with quantity");
+
+    m.def("make_mcscf_two_step", &make_mcscf_two_step, "state_weights_map"_a, "options"_a, "ints"_a,
```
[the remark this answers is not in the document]
fevangelista: I had a strange pybind11 issue, and moving it solved. I should investigate this a bit more because it sounds strange.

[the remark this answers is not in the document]
fevangelista: I moved it back where it belongs and things are OK. So I'll move it back where it was before.

forte/model.py
```diff
@@ -168,15 +168,65 @@ def state(self, charge: int, multiplicity: int, ms: float = None, sym: str = Non
...
         return StateInfo(na, nb, multiplicity, twice_ms, irrep, sym, gasmin, gasmax)
 
-    def ints(self, data, options):
+    def ints(self, data, options, df_basis_role: str = None):
+        """
+        This function prepares a ForteIntegral object from the data and options.
+
+        For density-fitted orbitals, we need to know the role of the auxiliary basis.
+        There are two options, SCF, used by codes that optimize MOs,
+        and CORR, used by methods to treat dynamical correlation.
```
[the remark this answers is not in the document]
fevangelista: This is a good point. Ideally we would like Forte to work with only one set of auxiliary function. This is here to support previous behavior. Some time ago we looked into the possibility of having mixed JK/RI basis sets that would work for both SCF and correlated methods, but it turns out that JK basis sets are usually better than RI ones. This is almost anecdotal, so it should be revisited in the future.

fevangelista:
More files have now been changed due to the renaming of a function. To answer your general question: we don't actually need to do a CI computation in between SCF and MCSCF. What the code expresses is the decoupling of the CI solver (which provides RDMs) and the MCSCF orbital optimizer. In practice, we do run the CI in between the two in the `run()` function of the active space solver. So we are wasting one CI cycle. I don't want to fix this issue at this point, because it would probably require a change in the way the MCSCF class works, and that's C++ code. But at least in this PR we are moving in the right direction, where we have decoupled the CI and orbital optimization parts of MCSCF (even on the C++ side).

fevangelista:
I reached a point in this PR where I need some help/opinion from @lcyyork. I am trying to figure out a good way to handle frozen orbitals in the MCSCF optimization. As far as I know, freezing core orbitals in MCSCF is rarely done, unless there are convergence issues connected to very low energy core orbitals. I guess that here there are at lease two strategies we can follow to enable freezing core/virtual orbital rotations in MCSCF: 

1. As the user to define only one MOSpaceInfo object with both frozen/restricted, but when passed to the MCSCF class, we do frozen core only  if the user indicates so explicitly (the rationale is that most times we do not want to freeze MCSCF).

2. Ask the user to define two MOSpaceInfo objects, one for the MCSCF computation, and one for a followup DSRG computation.

Now, I do favor solution #1, because most of the time we want to do a non-frozen MCSCF + frozen-DSRG computation, and therefore the input would become cumbersome. When I tried to implement it, I used one MOSpaceInfo object for the ActiveSpaceSolver (with frozen orbitals) and a different MOSpaceInfo object for the MCSCF (with only restricted, where I added up the frozen and restricted, and redefined the restricted space). The code ran, but the MCSCF energy was off. Therefore, I suspect there is something in the MCSCF code that does not work if the active space solver and the MSCF class use different MOSpaceInfo objects (even if they have the same active space). To summarize, I suppose it would be nice to find a good solution to handle the frozen MCSCF, and then it would be useful if you have any clue for solving the issue with the discrepancy.

fevangelista:
Another point to add to the discussion above: it looks like one of the limitations we face comes from the integral class. If we could work with one integral object that can support frozen and non-frozen orbitals at the same time we could avoid re-computing the integrals twice. This might be a bit of a overhaul, so for the moment I am inclined to skip supporting frozen core in MCSCF.

lcyyork:
@fevangelista May I ask how off was the energy? If they are off by several Hartree, I suspect it is because of the CI solver is `fci`. And the reason is the number of active electrons are not counted correctly. I think both `detci` and `cas` give the correct behavior.

JonathonMisiewicz:
> To answer your general question: we don't actually need to do a CI computation in between SCF and MCSCF. What the code expresses is the decoupling of the CI solver (which provides RDMs) and the MCSCF orbital optimizer. In practice, we do run the CI in between the two in the `run()` function of the active space solver. So we are wasting one CI cycle. I don't want to fix this issue at this point, because it would probably require a change in the way the MCSCF class works, and that's C++ code. But at least in this PR we are moving in the right direction, where we have decoupled the CI and orbital optimization parts of MCSCF (even on the C++ side).

I agree that worrying about that one CI iteration is premature optimization and outside the scope of this PR, but thinking about this has convinced me that I'm shaky on the computational graph aspects of the PR. I'm happy to have a call about this if you think it would be more efficient.

The ActiveSpaceSolver needs to get its integrals from the MCSCF solver, so why isn't there an arrow starting at MCSCF and going to ActiveSpaceSolver? This seems to count as "Take inputs from other nodes and perform a computational task."

docs/source/nb_api_intro.ipynb
```diff
@@ -78,36 +78,50 @@
     "```\n",
     "The output of this computation can be found in the `output.dat` file.\n",
     "However, the results of this computation are also stored in the HF object.\n",
-    "For example, the HF energy can be accessed via `hf.value('hf energy')`.\n"
+    "For example, the HF energy can be accessed via `hf.value('hf energy')`.\n",
+    "\n",
+    "To understand what happens when we setup this computation it is convenient to visualize it in the following way:\n",
```
JonathonMisiewicz, line 83:
```suggestion
    "To understand what happens when we setup this computation, it is convenient to visualize it in the following way:\n",
```

docs/source/nb_api_intro.ipynb
```diff
@@ -78,36 +78,50 @@
...
+    "\n",
+    "<center>\n",
+    "<img src=\"../source/img/api_intro-01.png\" alt=\"computation graph\" width=\"300\">\n",
+    "</center>\n",
+    "\n",
+    "This image represents the computation as a graph, where each step (setting up the input, running the HF computation) method is a node connected to the next step in a computation.\n",
```
JonathonMisiewicz, line 89:
```suggestion
    "This image represents the computation as a graph, where each step (setting up the input, running the HF computation) is a node, and two nodes are connected if the one on the right needs information from the one on the left.\n",
```

Also, what's the blue vs red significance?

docs/source/nb_api_intro.ipynb
```diff
@@ -78,36 +78,50 @@
...
+    "When we invoke the method `run()` on the Hartree–Fock object, the computation is executed, starting from the HF object backwards."
```
JonathonMisiewicz, line 90:
```suggestion
    "When we invoke the method `run()` on the Hartree–Fock object, it checks whether all its dependencies are run. (The input file doesn't need to do a computation, so it has run.) If not, it runs them recursively, until all dependencies have finished. The Hartree-Fock object can then start its own computation."
```

docs/source/nb_api_intro.ipynb
```diff
@@ -78,36 +78,50 @@
...
    "cell_type": "markdown",
    "id": "435ea455-ca6c-4f8b-8435-5762ef055352",
    "metadata": {},
    "source": [
-    "## FCI and CASCI\n",
+    "## Active space solvers: Full Configuration Interation (FCI), Complete Active Space CI (CASCI), and more\n",
     "\n",
     "Forte implements several solvers that diagonalize the Hamiltonian in a (small) space of orbitals, including FCI, selected CI methods, and generalized active space (GAS).\n",
-    "To perform one of these computations just pass an object that can provide molecular orbitals to an `ActiveSpaceSolver` object. For example, we can perform a CASCI computation on the same molecule as above by passing the `HF` node to an `ActiveSpaceSolver` node\n",
+    "To perform one of these computations just pass an object that can provide molecular orbitals to an `ActiveSpaceSolver` object. For example, we can perform a CASCI computation on the same molecule as above by passing the `HF` node to an `ActiveSpaceSolver` node, as shown in the following image\n",
```
JonathonMisiewicz, line 101:
```suggestion
    "To perform one of these computations, just pass an object that can provide molecular orbitals to an `ActiveSpaceSolver` object. For example, we can perform a CASCI computation on the same molecule as above by passing the `HF` node to an `ActiveSpaceSolver` node, as shown in the following image\n",
```

docs/source/nb_api_intro.ipynb
```diff
@@ -128,7 +142,45 @@
...
+   "source": [
+    "## Multiconfigurational SCF (MCSCF)\n",
+    "\n",
+    "To run an MCSCF computation in which we simultaneously optimize the orbitals and active space wave function, we use the MCSCF class.\n",
+    "This class requires as an input node an `ActiveSpaceSolver` object, and therefore, can be combine with any of the solvers implemented in Forte.\n",
+    "The following shows a scheme of a typical CASCI computation, using Hartree–Fock orbitals as initial guess for the MCSCF computation\n",
+    "<center>\n",
+    "<img src=\"../source/img/api_intro-03.png\" alt=\"computation graph\" style=\"width: 636px;\"/>\n",
+    "</center>\n",
+    "\n",
```
JonathonMisiewicz, line 162: Please add a comment to the effect of "The graph does _not_ mean that there is a single ActiveSpaceSolver iteration, and then the ActiveSpaceSolver is never touched again. It means that the MCSCF object requires the ActiveSpaceSolver as an input (to solve for active space wavefunctions using the current orbitals), and the ActiveSpaceSolver requires the HF object as an input (to determine orbital space dimensions)."

forte/base_classes/active_space_method.h
```diff
@@ -219,12 +219,6 @@ class ActiveSpaceMethod {
     void set_quite_mode(bool quiet);
 
   protected:
-    /// The list of active orbitals (absolute ordering)
-    std::vector<size_t> active_mo_;
```
JonathonMisiewicz, line 223: Three cheers for removing unused code.

forte/base_classes/active_space_solver.h
```diff
@@ -130,6 +130,13 @@ class ActiveSpaceSolver {
     /// Set if read wave function from file as initial guess
     void set_read_initial_guess(bool read_guess) { read_initial_guess_ = read_guess; }
 
+    /// @return the the label of the ActiveSpaceMethod used by this solver
+    std::string method() const;
+
+    std::shared_ptr<MOSpaceInfo> mo_space_info() const;
```
JonathonMisiewicz, line 136: If we don't want this memory address to change, we might want to mark this ptr const. Same below.

forte/base_classes/active_space_solver.h
```diff
@@ -130,6 +130,13 @@ class ActiveSpaceSolver {
     /// Set if read wave function from file as initial guess
     void set_read_initial_guess(bool read_guess) { read_initial_guess_ = read_guess; }
 
+    /// @return the the label of the ActiveSpaceMethod used by this solver
+    std::string method() const;
```
JonathonMisiewicz, line 134: Is there a reason not to return const &?

forte/base_classes/mo_space_info.cc
```diff
@@ -51,6 +51,18 @@ MOSpaceInfo::MOSpaceInfo(const psi::Dimension& nmopi, const std::string& point_g
     }
 }
 
+size_t MOSpaceInfo::nirrep() const { return nirrep_; }
+
+const std::string& MOSpaceInfo::irrep_label(size_t h) const { return symmetry_.irrep_label(h); }
+
+std::string MOSpaceInfo::point_group_label() const { return symmetry_.point_group_label(); }
```
JonathonMisiewicz, line 58: Is there a reason not to return const &?

forte/casscf/mcscf_2step.cc
```diff
@@ -91,7 +92,7 @@ void MCSCF_2STEP::read_options() {
 
     orb_type_redundant_ = options_->get_str("CASSCF_FINAL_ORBITAL");
 
-    ci_type_ = options_->get_str("CASSCF_CI_SOLVER");
+    ci_type_ = active_space_solver_->method();
```
JonathonMisiewicz, line 95: Hooray for fewer options dependencies.

forte/fci/string_lists.h
```diff
@@ -119,20 +119,18 @@ class StringLists {
...
-    int nirrep() const { return nirrep_; }
-    size_t ncmo() const { return ncmo_; }
-    std::vector<size_t> cmo_to_mo() const { return cmo_to_mo_; }
-    std::vector<size_t> fomo_to_mo() const { return fomo_to_mo_; }
-    psi::Dimension cmopi() const { return cmopi_; }
-    std::vector<size_t> cmopi_offset() const { return cmopi_offset_; }
     size_t nb() const { return nb_; }
+    int nirrep() const { return nirrep_; }
+    size_t nmo() const { return nmo_; }
+    psi::Dimension mopi() const { return mopi_; }
```
JonathonMisiewicz, line 132: Again, is this an occasion for const &? Same on the line below.

forte/sci/detci.cc
```diff
@@ -123,7 +123,8 @@ void DETCI::build_determinant_space() {
     if (size == 0) {
         outfile->Printf("\n  There is no determinant matching the conditions!");
         outfile->Printf("\n  Please check the input (symmetry, multiplicity, etc.)!");
-        throw std::runtime_error("No determinant matching the conditions!");
+        throw std::runtime_error(
```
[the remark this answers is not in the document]
JonathonMisiewicz: This comment still applies.

forte/solvers/factory.py
```diff
@@ -5,8 +5,21 @@
...
 
-def solver_factory(molecule, basis, int_type=None, scf_aux_basis=None, corr_aux_basis=None):
-    """A factory to build a basic solver object"""
+def input_factory(molecule, basis, int_type=None, jkfit_aux_basis=None, rifit_aux_basis=None):
+    """
+    A factory to build an input solver object
+
+    Parameters
+    ----------
+    basis: str or Basis
```
JonathonMisiewicz, line 14: Document molecule argument.

forte/solvers/feature.py
```diff
@@ -0,0 +1,16 @@
+from enum import Enum, auto
+
+
+class Feature(Enum):
+    """
+    This enum class is used to store all types of features
```
JonathonMisiewicz, line 6: Document what a "feature" is. I think you are the person in this group most acquainted with computational graph technical jargon.

forte/solvers/mcscf.py
```diff
@@ -0,0 +1,132 @@
...
+from forte.solvers.solver import Solver
+
+from forte.forte import ForteOptions
+from forte import forte_options
+from forte.forte import make_mcscf_two_step, make_mo_space_info_from_map
+
+
+class MCSCF(Solver):
+    """
+    A class to perform orbital optimization
```
JonathonMisiewicz, line 13: This docstring has me thinking: can I use this class to do OMP2, or is it just a class to do orbital optimized determinant-based computations?

tests/pytest-methods/casscf/test_casscf-7.py
```diff
@@ -0,0 +1,75 @@
...
+    mcscf = MCSCF(fci)
+    mcscf.run()
+
+    assert mcscf.value('mcscf energy')[state] == pytest.approx([ref_mcscf_energy], 1.0e-10)
+
+
+if __name__ == "__main__":
+    test_casscf_7()
+
+# import forte
```
JonathonMisiewicz, line 34: Is there a reason to keep this around?

JonathonMisiewicz (commented):
I've given this a relatively quick review. The actual MCSCF part looks fine to me, though York is much more qualified to give feedback there. My bigger concern is getting the computational graphs explained in a way that's more explicit.

JonathonMisiewicz:
Please ping me if this PR needs my attention - I'm unclear of the PR status at this stage.

PR #261, opened 2021-06-29
AVAS π Orbital Selection

lcyyork:
## Description
This PR add the support of pruning the subspace p orbitals to those only orthogonal to the given planes.

There are a few motivations of doing so.
- The xyz frame of atomic orbitals may not be aligned to the molecular xyz frame such that `pz` orbitals are not perpendicular to the desired plane.
- There may be multiple π subsystems in a molecule and one cannot make the molecular xyz frame appropriate to all subsystems.
- For realistic systems, we mostly likely encounter *approximate* planar structure and we have to define an averaged plane.

#### A simple example of formaldehyde (test `avas-7`, modified from `avas-1`).

The xyz coordinate is obtained by rotating that from test case `avas-1`. Now the plane normal is (1.0, 1.0, 1.0) / √3.
```
C        0.346146295209737    0.126698337466632   -0.472844632676369
O       -0.346053592352928   -0.126664405871036    0.472717998223964
H        1.227335215970831   -0.489581944167415   -0.737753271803415
H        0.143281782803090    0.991262584630455   -1.134544367433545
```

By specifying the following options, `AOSubspace` class is able to calculate the plane normal and determine the subspace AOs.
```
# have to include all p orbitals
subspace ["C(2p)", "O(2P)"]

# a list of planes defined by a list of atoms
subspace_pi_planes [["C", "O", "H"]]
```
The subspace orbitals are defined by a linear combination of the given `p` orbitals.
The output from the above example reads as
```
  ==> List of Planes Requested <==

    Plane  1
             C         O         H

  ==> List of Subspace Orbitals Requested <==
         C(2P)         O(2P)
  NOTE: Subspace orbitals may be truncated based on requested planes!

  ==> AO Basis Set Selected By Subspace <==

    =======================================
      AO  Atom  Label     Type  Coefficient
    ---------------------------------------
       2     1     C1      2PZ   5.7735E-01
       3     1     C1      2PX   5.7735E-01
       4     1     C1      2PY   5.7735E-01
       7     2     O1      2PZ   5.7735E-01
       8     2     O1      2PX   5.7735E-01
       9     2     O1      2PY   5.7735E-01
    ---------------------------------------
    Number of subspace orbitals:          2
    =======================================
```

One could use double-π active space by including `"C(3p)"` and `"O(3p)"`.

#### A more advanced example of C20 fullerene (check test case avas-8).

We need to find all 12 planes of C20 and pass to `subspace_pi_planes`.
Check the molecule and we find each atom is shared by three planes.
Note that the molecule is in C1 symmetry due to Jahn-Teller effect.

geom

The following code finds all the planes and stored as a list of list, which is then passed to `subspace_pi_planes` via Psi4 option interface.
```
# figure out planes
import numpy as np
from collections import defaultdict
from itertools import combinations

# grab molecule xyz
mol = psi4.core.get_active_molecule()
xyz = np.array(mol.geometry())  # in Bohr
n_atoms = xyz.shape[0]

# compute bond vectors
bond_vectors = {}
for i in range(n_atoms):
    for j in range(i + 1, n_atoms):
        bond = xyz[i] - xyz[j]
        bond_vectors[(i, j)] = bond
        bond_vectors[(j, i)] = bond

# compute bond lengths
bond_lengths = defaultdict(dict)
for p, b in bond_vectors.items():
    i, j = p
    r = np.linalg.norm(b)
    bond_lengths[i][j] = r
    bond_lengths[j][i] = r

# keep only the nearest and second nearest neighbors
neighbors = {i: sorted(p.keys(), key=lambda x: p[x])[:9] for i, p in bond_lengths.items()}

# find the planes
planes = []
for i in range(n_atoms):
    nearest1 = neighbors[i][:3]
    nearest2 = neighbors[i][3:]
    for j, k in combinations(nearest1, 2):
        if j < i or k < i:
            continue
        j_nearest = [m for m in neighbors[j][:3] if m != i and m in neighbors[k][3:]][0]
        k_nearest = [m for m in neighbors[k][:3] if m != i and m in neighbors[j][3:]][0]
        plane = [f"C{m + 1}" for m in [i, j, k, j_nearest, k_nearest]]
        planes.append(plane)

set forte {
  job_type           none
  subspace           ["C(2p)"]  # must include all p orbitals
  subspace_pi_planes $planes  # a list of planes
  avas               true
  avas_diagonalize   true
  avas_cutoff        0.5
}
Ezero, wfn = energy('forte', ref_wfn=wfn, return_wfn=True)
```
In the output file, we should get
```
  ==> List of Planes Requested <==

    Plane  1
          C1     C12      C3      C9      C6
    Plane  2
          C1     C12      C2     C13     C15
    Plane  3
          C1      C3      C2      C4      C5
    Plane  4
          C2     C15      C5     C16     C20
    Plane  5
          C3      C6      C4      C7      C8
    Plane  6
          C4      C5      C8     C20     C19
    Plane  7
          C6      C9      C7     C10     C11
    Plane  8
          C7     C11      C8     C18     C19
    Plane  9
          C9     C10     C12     C14     C13
    Plane 10
         C10     C14     C11     C17     C18
    Plane 11
         C13     C14     C15     C17     C16
    Plane 12
         C16     C17     C20     C18     C19

  ==> List of Subspace Orbitals Requested <==
         C(2P)
  NOTE: Subspace orbitals may be truncated based on requested planes!

  ==> AO Basis Set Selected By Subspace <==

    =======================================
      AO  Atom  Label     Type  Coefficient
    ---------------------------------------
       2     1     C1      2PZ  -4.7137E-01
       3     1     C1      2PX   8.1782E-01
       4     1     C1      2PY  -3.3013E-01
       7     2     C2      2PZ  -2.3524E-01
       8     2     C2      2PX   9.1378E-01
       9     2     C2      2PY   3.3117E-01
      12     3     C3      2PZ  -9.1169E-01
      13     3     C3      2PX   2.4813E-01
      14     3     C3      2PY  -3.2751E-01
      17     4     C4      2PZ  -9.4252E-01
      18     4     C4      2PX  -1.2479E-02
      19     4     C4      2PY   3.3392E-01
      22     5     C5      2PZ  -5.3279E-01
      23     5     C5      2PX   4.1031E-01
      24     5     C5      2PY   7.4012E-01
      27     6     C6      2PZ  -6.1479E-01
      28     6     C6      2PX  -2.6722E-01
      29     6     C6      2PY  -7.4204E-01
      32     7     C7      2PZ  -4.6484E-01
      33     7     C7      2PX  -8.2341E-01
      34     7     C7      2PY  -3.2545E-01
      37     8     C8      2PZ  -6.7400E-01
      38     8     C8      2PX  -6.6405E-01
      39     8     C8      2PY   3.2366E-01
      42     9     C9      2PZ   5.6768E-03
      43     9     C9      2PX   1.0583E-03
      44     9     C9      2PY  -9.9998E-01
      47    10    C10      2PZ   5.2790E-01
      48    10    C10      2PX  -4.1076E-01
      49    10    C10      2PY  -7.4337E-01
      52    11    C11      2PZ   2.3128E-01
      53    11    C11      2PX  -9.1234E-01
      54    11    C11      2PY  -3.3785E-01
      57    12    C12      2PZ   8.6568E-02
      58    12    C12      2PX   6.6569E-01
      59    12    C12      2PY  -7.4119E-01
      62    13    C13      2PZ   6.6603E-01
      63    13    C13      2PX   6.6964E-01
      64    13    C13      2PY  -3.2861E-01
      67    14    C14      2PZ   9.4233E-01
      68    14    C14      2PX   6.3845E-03
      69    14    C14      2PY  -3.3464E-01
      72    15    C15      2PZ   4.6576E-01
      73    15    C15      2PX   8.1931E-01
      74    15    C15      2PY   3.3437E-01
      77    16    C16      2PZ   6.1817E-01
      78    16    C16      2PX   2.6383E-01
      79    16    C16      2PY   7.4044E-01
      82    17    C17      2PZ   9.1077E-01
      83    17    C17      2PX  -2.4235E-01
      84    17    C17      2PY   3.3430E-01
      87    18    C18      2PZ   4.7339E-01
      88    18    C18      2PX  -8.1869E-01
      89    18    C18      2PY   3.2505E-01
      92    19    C19      2PZ  -9.9132E-02
      93    19    C19      2PX  -6.6468E-01
      94    19    C19      2PY   7.4052E-01
      97    20    C20      2PZ   2.1938E-03
      98    20    C20      2PX   7.1143E-03
      99    20    C20      2PY   9.9997E-01
    ---------------------------------------
    Number of subspace orbitals:         20
    =======================================
```
A sample output for the AVAS procedure would be
```
  Diagonalizing the doubly occupied projected overlap matrix ... Done
  Diagonalizing the unoccupied projected overlap matrix ........ Done
  Sum of eigenvalues: 19.93760447
  AVAS covers 96.89% of the subspace.

  ==> AVAS MOs Information <==

    ---------------------
                        A
    ---------------------
    DOCC INACTIVE      50
    DOCC ACTIVE        10
    SOCC ACTIVE         0
    UOCC ACTIVE        10
    UOCC INACTIVE     210
    ---------------------
    RESTRICTED_DOCC    50
    ACTIVE             20
    RESTRICTED_UOCC   210
    ---------------------

  ==> Atomic Valence MOs (Active Marked by *) <==

    ===============================
     Irrep    MO  Occ.  <phi|P|phi>
    -------------------------------
    *   A      0    2      0.987854
    *   A      1    2      0.978370
    *   A      2    2      0.978231
    *   A      3    2      0.977779
    *   A      4    2      0.949803
    *   A      5    2      0.949060
    *   A      6    2      0.944495
    *   A      7    2      0.942943
    *   A      8    2      0.941567
    *   A      9    2      0.921297
        A     10    2      0.047860
        A     11    2      0.042087
        A     12    2      0.039389
        A     13    2      0.037573
        A     14    2      0.011928
        A     15    2      0.011261
        A     16    2      0.010241
        A     17    2      0.007468
        A     18    2      0.007059
        A     19    2      0.006744
    *   A     60    0      0.989979
    *   A     61    0      0.989659
    *   A     62    0      0.989237
    *   A     63    0      0.986610
    *   A     64    0      0.985585
    *   A     65    0      0.984927
    *   A     66    0      0.959397
    *   A     67    0      0.957579
    *   A     68    0      0.954881
    *   A     69    0      0.949105
        A     70    0      0.075555
        A     71    0      0.055008
        A     72    0      0.053634
        A     73    0      0.052075
        A     74    0      0.047529
        A     75    0      0.046780
        A     76    0      0.019356
        A     77    0      0.018935
        A     78    0      0.018818
        A     79    0      0.009945
    ===============================
```
The first active orbital is virus looking, which I found interesting.
Psi_a_51_51-A

## TODOs
- A Python version of `AOSubspace`?

## Checklist
- [x] Documented source code
- [x] Ready to review.
- [x] Ready to go!


lcyyork:
@fevangelista  I no longer use the old variables (e.g., `subspace_`) and functions. Should I delete them or do you want to keep them?

fevangelista:
@lcyyork: I need a code refresh to answer your question. But in the mean time I was wondering: can you automate the plane detection by looking for the 3 closest atoms to the center you select? The current way to provide input seems open to ambiguity and would a bit limited in terms of applications.

lcyyork:
@fevangelista Please let me know what is the "ambiguity" you have in mind. In my opinion, the most general approach is let the user decide what they want to do. So I am not sure why it is "limited in terms of applications". A snippet of how your option looks like / works will be helpful.

I can implement the nearest three-atom method but it may not be ideal. In fact, that may only be useful for fullerene related molecules. For example,
- Assume a planar molecule consisting of a chain of carbons ended with CH2 on both ends (e.g., H2C-C-C-C-C-C-C-CH2). It is unclear what the plane is for the inner C atoms.
- For a porphyrin molecule slightly distorted due to ligands, what if the user want the plane averaged over all porphyrin atoms?
- For porphyrin, hydrogen atoms must be included for edge carbon/nitrogen atoms for this scheme to work.

For fullerenes, if the user pass in the correct planes, the current interface is still able to correctly build the subspace orbitals. This does not mean the user have to input all the planes by hand but he/she can write some code and use the Psi4 option interface to pass in a python list, which is prepared by the user. For example, I add an example for C20 molecule and here is the lowest occupied π orbital.
Psi_a_51_51-A

README.md
```diff
@@ -176,3 +176,15 @@ or add the option
   -DENABLE_CODECOV=ON
   ```
 if compiling with CMake.
+
+### Build the manual locally
+
+The documentation can be built locally using `sphinx`.
+Additionally, `sphinx_rtd_theme` and `nbsphinx` modules are required.
```
fevangelista, line 183:
```suggestion
Additionally, the `sphinx_rtd_theme` and `nbsphinx` modules are required.
```

README.md
```diff
@@ -176,3 +176,15 @@ or add the option
...
+  ```
+  cd <homedir>/src/forte/docs
```
fevangelista, line 185:
```suggestion
from the forte source directory (e.g., $HOME/src/forte) change to the directory `docs` and call `make html`:
  cd docs
```

forte/orbital-helpers/aosubspace.h
```diff
@@ -137,48 +142,29 @@ class AOSubspace {
     // Constructor with list of subspaces
```
fevangelista, line 142: Please remove the first constructor and give the argument atom_normals an empty map. It's easy to specify this in pybind11, there are other example in the various api files.

forte/orbital-helpers/aosubspace.h
```diff
@@ -137,48 +142,29 @@ class AOSubspace {
...
     AOSubspace(std::vector<std::string> subspace_str, std::shared_ptr<psi::Molecule> molecule,
```
fevangelista, line 143:
```suggestion
```

forte/orbital-helpers/aosubspace.h
```diff
@@ -137,48 +142,29 @@ class AOSubspace {
...
                std::shared_ptr<psi::BasisSet> basis);
```
fevangelista, line 144:
```suggestion
```

forte/orbital-helpers/aosubspace.h
```diff
@@ -137,48 +142,29 @@ class AOSubspace {
...
+    /// Constructor using the atom normals
+    /// an atom normal is a 3D vector onto which p orbitals are projected:
+    /// px, py, pz (3 orbitals) -> nx px + ny py + nz pz (1 orbital)
+    AOSubspace(std::vector<std::string> subspace_str, std::shared_ptr<psi::Molecule> molecule,
+               std::shared_ptr<psi::BasisSet> basis,
+               std::map<std::pair<int, int>, psi::Vector3> atom_normals, bool debug_mode = false);
```
fevangelista, line 150:
```suggestion
               std::map<std::pair<int, int>, psi::Vector3> atom_normals = std::map<std::pair<int, int>, psi::Vector3>(), bool debug_mode = false);
```

forte/orbital-helpers/aosubspace.h
```diff
@@ -137,48 +142,29 @@ class AOSubspace {
     // Constructor with list of subspaces
     AOSubspace(std::vector<std::string> subspace_str, std::shared_ptr<psi::Molecule> molecule,
                std::shared_ptr<psi::BasisSet> basis);
+    /// Constructor using the atom normals
+    /// an atom normal is a 3D vector onto which p orbitals are projected:
+    /// px, py, pz (3 orbitals) -> nx px + ny py + nz pz (1 orbital)
```
fevangelista, line 147: This function needs to document the input better. What are the two integers passed in the pair? Here you should mention that this is an optional value that can be used to define a subspace of p orbitals aligned to the atom normals.

fevangelista (changes requested):
Thanks for changing the code to support normals! Here are some initial comments for you to work on. I'd like to do a second more careful pass later.

README.md
lcyyork: I remove "$HOME/" from the suggestion because it is not the way we describe the directory at the beginning of this README.

I wonder if I should move this part inside `docs`. Maybe that fits better.

forte/orbital-helpers/aosubspace.h
lcyyork: Let me do that in initialization list.

forte/orbital-helpers/aosubspace.h
lcyyork: OK, I misunderstood your previous comment. Changed.

forte/orbital-helpers/aosubspace.cc
```diff
@@ -62,84 +65,74 @@ std::vector<std::string> mysplit(const std::string& input, const std::string& re
...
-    parse_basis_set();
-    parse_subspace();
-}
-
 void AOSubspace::startup() {
-    //    outfile->Printf("  ---------------------------------------\n");
-    //    outfile->Printf("    Atomic Orbital Subspace\n");
-    //    outfile->Printf("    written by Francesco A. Evangelista\n");
-    //    outfile->Printf("  ---------------------------------------\n");
-
```
lcyyork, line 142: I wonder `lm_labels_cartesian_ ` should be removed because it is not used. We could test the `min_basis_` to make sure it is `pure_am`.

forte/orbital-helpers/aosubspace.cc
```diff
@@ -148,173 +141,135 @@ void AOSubspace::startup() {
...
-                           {"DZ2", "DXZ", "DYZ", "DX2-Y2", "DXY"},
-                           {"FZ3", "FXZ2", "FYZ2", "FZX2-ZY2", "FXYZ", "FX3-3XY2", "F3X2Y-Y3"},
-                           {"G1", "G2", "G3", "G4", "G5", "G6", "G7", "G8", "G9"},
-                           {"H1", "H2", "H3", "H4", "H5", "H6", "H7", "H8", "H9", "H10", "H11"}};
+    lm_labels_spherical_ = {{"S"},
+                            {"PZ", "PX", "PY"},
+                            {"DZ2", "DXZ", "DYZ", "DX2-Y2", "DXY"},
+                            {"FZ3", "FXZ2", "FYZ2", "FZX2-ZY2", "FXYZ", "FX3-3XY2", "F3X2Y-Y3"},
+                            {"G1", "G2", "G3", "G4", "G5", "G6", "G7", "G8", "G9"},
+                            {"H1", "H2", "H3", "H4", "H5", "H6", "H7", "H8", "H9", "H10", "H11"}};
```
lcyyork, line 149: @fevangelista Do you prefer putting these labels here or in the header? I don't think this should be changed anyway.

forte/orbital-helpers/aosubspace.h
lcyyork: This is in favor of how `atom_to_aos_` is accessed. I added more comments.

README.md
fevangelista: I would move it to the docs. Note that if you want to modify docs in notebooks, you will have to change the corresponding ipynb first, then call a script in the docs folder (all of this is documented).

forte/orbital-helpers/aosubspace.h
fevangelista: No, I don't think we need to expose this. With one constructor, we would probably will just have to modify the interface of make_aosubspace_projector to deal with the optional variable.

forte/orbital-helpers/aosubspace.cc
fevangelista: I suppose the idea was to support both 5D and 6D type of functions, but that never happened. So I think it's a good idea to remove cartesians and check for pure_am. I don't remember, but are pure_am basis sets still pure cartesians, like px, py, pz?

README.md
lcyyork: Moved.

docs/source/methods/avas.rst
```diff
@@ -168,6 +168,197 @@ We then use this guess of active orbitals to compute the CASCI energy:::
...
+    avas_cutoff         0.5
+    minao_basis         cc-pvtz-minao
+    subspace            ["Fe(3d)","C6-25(2p)","N(2p)","S(3p)","C1-3(2p)"]
+    subspace_pi_planes  [["Fe","C6-25","N3-6"], ["N1-2","C1-3"]]
+  }
+
+The AVAS output selects exactly 37 orbitals we wanted. ::
+
+  Sum of eigenvalues: 36.98863914
+  AVAS covers 96.59% of the subspace.
```
rwashi690, line 263: Does this come from the overlap matrix? If so SVD would be good to compute and get the average of the singular values

tests/methods/avas-7/input.dat
```diff
@@ -0,0 +1,41 @@
...
+set {
+  basis cc-pvdz
+  reference rhf
+  scf_type pk
+  e_convergence 12
+}
+
+set forte {
+  subspace ["C(2p)","O(2P)"]  # must include all p orbitals
+  subspace_pi_planes [["C","O","H"]]  # a list of planes
```
rwashi690, line 28: Here would be a place to specify what plane the orbitals are on. Similar to before like on the xz plane

tests/methods/avas-8/input.dat
```diff
@@ -0,0 +1,114 @@
+# Test the pi orbital selection of AOSubspace on overlapped planes C20 molecule
+
+import forte
+
+refscf = -756.50604655444420
+reffci = -113.911667467206257
+
+molecule C20{
+# grabed from https://nanotube.msu.edu/fullerene/fullerene.php?C=20
+# C1 symmetry due to Jahn-Teller distortion
```
rwashi690, line 10: To remind people state that fullerene molecule C20 is in cage form and usually has an icosahedral geometry with a Ih point group

tests/methods/avas-8/input.dat
```diff
@@ -0,0 +1,114 @@
+# Test the pi orbital selection of AOSubspace on overlapped planes C20 molecule
+
```
rwashi690, line 2: Add in the description that this uses forte to find the planes of a fullerene molecule in forte

tests/methods/avas-8/input.dat
```diff
@@ -0,0 +1,114 @@
...
+  basis         cc-pvdz
+  df_basis_scf  cc-pvdz-jkfit
+  reference     rhf
+  scf_type      df
+  d_convergence 10
+}
+Escf, wfn = energy('scf', return_wfn=True)
+compare_values(refscf, variable("CURRENT ENERGY"), 8, "SCF energy")
+
+# figure out planes
```
rwashi690, line 43: Add -> "Here we start to find the planes of fullerene in cage form and import some necessary python modules to complete this calculation"

tests/methods/avas-8/input.dat
```diff
@@ -0,0 +1,114 @@
...
+import numpy as np
+from collections import defaultdict
+from itertools import combinations
+
```
rwashi690, line 47: First we grab the geometry of fullerene in cartesian coordinates and import it into the python numpy  module

tests/methods/avas-8/input.dat
```diff
@@ -0,0 +1,114 @@
...
+# grab molecule xyz
+mol = psi4.core.get_active_molecule()
+xyz = np.array(mol.geometry())  # in Bohr
+n_atoms = xyz.shape[0]
+
+# compute bond vectors
```
rwashi690, line 53: compute the bond vectors of fullerene

tests/methods/avas-8/input.dat
```diff
@@ -0,0 +1,114 @@
...
+bond_vectors = {}
+for i in range(n_atoms):
+    for j in range(i + 1, n_atoms):
+        bond = xyz[i] - xyz[j]
+        bond_vectors[(i, j)] = bond
+        bond_vectors[(j, i)] = bond
+
+# compute bond lengths
```
rwashi690, line 61: compute bond lengths of fullerene

tests/methods/avas-8/input.dat
```diff
@@ -0,0 +1,114 @@
...
+bond_lengths = defaultdict(dict)
+for p, b in bond_vectors.items():
+    i, j = p
+    r = np.linalg.norm(b)
+    bond_lengths[i][j] = r
+    bond_lengths[j][i] = r
+
+# keep only the nearest and second nearest neighbors
```
rwashi690, line 69: specify why you keep only the nearest and second nearest neighbors

tests/methods/avas-8/input.dat
```diff
@@ -0,0 +1,114 @@
...
+neighbors = {i: sorted(p.keys(), key=lambda x: p[x])[:9] for i, p in bond_lengths.items()}
+
+# find the planes
```
rwashi690, line 72: Add to comment -> then we utilize the two nearest neighbors to calculate the planes occupied by the 2pz orbitals (defined by a linear combination of all 2p orbitals) of the 20 carbon atoms in fullerene

tests/methods/avas-8/input.dat
```diff
@@ -0,0 +1,114 @@
...
+            continue
+        j_nearest = [m for m in neighbors[j][:3] if m != i and m in neighbors[k][3:]][0]
+        k_nearest = [m for m in neighbors[k][:3] if m != i and m in neighbors[j][3:]][0]
+        plane = [f"C{m + 1}" for m in [i, j, k, j_nearest, k_nearest]]
+        planes.append(plane)
+
+set forte {
+  job_type           none
+  subspace           ["C(2p)"]  # must include all p orbitals
+  subspace_pi_planes $planes  # a list of planes
```
rwashi690, line 88: add to comment "list the planes from the function we determined via the cartesian geometry"

tests/methods/avas-8/input.dat
```diff
@@ -0,0 +1,114 @@
...
+  avas               true
+  avas_diagonalize   true
+  avas_cutoff        0.5
+}
+Ezero, wfn = energy('forte', ref_wfn=wfn, return_wfn=True)
+
+# fix orbital phase
```
rwashi690, line 95: Explain in more details what you mean by fix orbital phase.

tests/methods/avas-8/input.dat
```diff
@@ -0,0 +1,114 @@
...
+Ca = wfn.Ca()
+for i in range(Ca.cols()):
+    if Ca.get(0, i) < 0:
+        Ca.scale_column(0, i, -1.0)
+
+# test only the core and active orbitals
```
rwashi690, line 101: Here more explanation

tests/methods/avas-8/input.dat
```diff
@@ -0,0 +1,114 @@
...
+set forte {
+  job_type           none
+  subspace           ["C(2p)"]  # must include all p orbitals
+  subspace_pi_planes $planes  # a list of planes
+  avas               true
+  avas_diagonalize   true
+  avas_cutoff        0.5
+}
+Ezero, wfn = energy('forte', ref_wfn=wfn, return_wfn=True)
+
```
rwashi690, line 94: Before continuing with the next section specify what the next block of code aims to determine

rwashi690: So at the bottom I realized that you are comparing previous calculation to reference orbitals. Add that up here before continuing

tests/methods/tests.yaml
```diff
@@ -55,11 +55,14 @@ avas:
    short:
```
rwashi690, line 55: When learning how to use forte to be honest this set-up of avas-3, avas-5 didn't help me figure out what I was looking for. In future as a group we need to figure out a better way to indicate what the test case is before people open the files. Perhaps we should have a short description of each test case in the manual. I know that will be pretty time consuming but I am willing to help with this task

docs/source/methods/avas.rst
```diff
@@ -168,6 +168,197 @@ We then use this guess of active orbitals to compute the CASCI energy:::
...
+where each plane is given by a list of atoms' expressions.
+Some valid expressions would be: ::
+
+  - [['C', 'H', 'O']]  # only one plane consisting all C, H, and O atoms of the molecule.
+  - [['C1-6'], ['N1-2', 'C9-11']]  # plane 1 with the first six C atoms of the molecule,
+                                   # plane 2 with C9, C10, C11, N1 and N2 atoms.
+  - [['C1-4'], ['C1-2', 'C5-6']]:  # plane 1 with the first four C atoms of the molecule,
+                                   # plane 2 with C1, C2, C5, and C6 atoms.
+                                   # Two planes share C1 and C2!
+
```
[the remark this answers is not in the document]
lcyyork: @rwashi690  I think you miss the point. Let me rewrite the manual.

This part is just to illustrate the syntax of that keyword.

What I want to achieve is very simple. So suppose using the provided formaldehyde geometry, which is not lying in any of the xy, zx, or yz plane. Now I want AVAS select the pi orbitals of the molecule, which consist of the C and O 2p orbitals that are perpendicular to the plane. Now how to choose these two p orbitals? They are not aligned to x, y, nor z direction. You could try to use ["C(2px)", "O(2px)"] and see what AVAS yields.

There are two ways for this particular case. We could try to rotate the formaldehyde molecule such that the molecular plane is located at xy, xz, or yz plane, and the 2p orbitals are simply aligned with the other direction. For example, if the molecule is at yz plane, the subspace orbitals would be ["C(2px)", "O(2px)"].

The other approach is not to move the molecule but to project all three 2p orbitals onto the plane normal. Suppose the plane normal is (nx, ny, nz). The 2p orbital perpendicular to the plane is simply nx * 2px + ny * 2py + nz * 2pz. This is what I mean px, py, pz orbitals goes to 'pz' orbital that is perpendicular to the plane. Note the quotation mark.

Which approach is more general? I must say it is the second because there might be multiple pi systems in a molecule.

tests/methods/avas-8/input.dat
lcyyork: I think the C20 cage is usually in C1 but mentioning this is a cage is a good point.

docs/source/methods/avas.rst
```diff
@@ -168,6 +168,197 @@ We then use this guess of active orbitals to compute the CASCI energy:::
...
+The following input snippet selects all 3d orbitals of Fe, 3p orbitals of S,
+and all 'pz' orbitals of porphyrin and imidazole rings. ::
+
+  set forte {
+    avas                true
+    avas_diagonalize    true
+    avas_cutoff         0.5
+    minao_basis         cc-pvtz-minao
+    subspace            ["Fe(3d)","C6-25(2p)","N(2p)","S(3p)","C1-3(2p)"]
+    subspace_pi_planes  [["Fe","C6-25","N3-6"], ["N1-2","C1-3"]]
```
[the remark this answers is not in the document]
lcyyork: It is not possible to make such claim. Neither of them locate on xz or yz plane.

docs/source/methods/avas.rst
```diff
@@ -168,6 +168,197 @@ We then use this guess of active orbitals to compute the CASCI energy:::
...
+  C        0.346146295209737    0.126698337466632   -0.472844632676369
+  O       -0.346053592352928   -0.126664405871036    0.472717998223964
+  H        1.227335215970831   -0.489581944167415   -0.737753271803415
+  H        0.143281782803090    0.991262584630455   -1.134544367433545
+  noreorient
+  }
+
+  set forte {
+    subspace           ["C(2p)", "O(2p)"]  # must include all p orbitals!
+    subspace_pi_planes [["C", "O", "H"]]   # only one plane, defined by all C, O and H atoms
```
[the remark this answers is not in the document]
lcyyork: The definition of plane has nothing to do orbital orbitals, but only atoms.

docs/source/methods/avas.rst
```diff
@@ -168,6 +168,197 @@ We then use this guess of active orbitals to compute the CASCI energy:::
...
+  0 1
+  C        0.346146295209737    0.126698337466632   -0.472844632676369
+  O       -0.346053592352928   -0.126664405871036    0.472717998223964
+  H        1.227335215970831   -0.489581944167415   -0.737753271803415
+  H        0.143281782803090    0.991262584630455   -1.134544367433545
+  noreorient
+  }
+
+  set forte {
+    subspace           ["C(2p)", "O(2p)"]  # must include all p orbitals!
```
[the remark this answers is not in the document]
lcyyork: See if my previous comment make sense.

docs/source/methods/avas.rst
lcyyork: This is the percentage of AVAS selected orbitals over the total subspace orbitals. The SVD values are actually printed afterwards.

tests/methods/avas-7/input.dat
lcyyork: The plane function is 0 = (x - Cx) + (y - Cy) + (z - Cz) = (x - Ox) + (y - Oy) + (z - Oz) = (x - H1x) + (y - H1y) + (z - H1z) = (x - H2x) + (y - H2y) + (z - H2z) because the plane normal is (1, 1, 1).

tests/methods/avas-8/input.dat
lcyyork: The planes are figured by the user, not Forte.

tests/methods/tests.yaml
lcyyork: I guess it depends on what kind of information should be included in test cases and what should be included in the manual. The DSRG manual contains a complete list of test cases, but the description of each test case is nowhere near complete.

docs/README.md
```diff
@@ -0,0 +1,10 @@
+### Build the manual locally
+
+If you are temporarily offline but would like to check the manual,
+you might want to compile the manual locally.
+This can be done using `sphinx`.
+Additionally, the `sphinx_rtd_theme` and `nbsphinx` modules are required.
+Since you already find this README, simply call `make html` in the current directory.
```
fevangelista, line 7:
```suggestion
Since you already found this README file, simply call `make html` in the current directory.
```

docs/README.md
```diff
@@ -0,0 +1,10 @@
+### Build the manual locally
```
fevangelista, line 1:
```suggestion
### Building Forte's manual locally
```

docs/source/methods/avas.rst
```diff
@@ -66,7 +66,7 @@ Input Example
 ^^^^^^^^^^^^^
 
 In this example, we perform the AVAS procedure on formaldehyde
```
fevangelista, line 68:
```suggestion
In this example, we use AVAS to find an active space for formaldehyde that spans the :math:`2p_x` orbitals of the C and O atoms,
```

docs/source/methods/avas.rst
```diff
@@ -168,6 +168,241 @@ We then use this guess of active orbitals to compute the CASCI energy:::
 For more examples, see :code:`avas-1` to :code:`avas-6` in the :code:`tests/methods` folder.
 In particular, :code:`avas-6` is a practical example on ferrocene.
 
+Defining the molecular plane for π orbitals
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Molecular systems with conjugated π bonds generally process planar geometries.
```
fevangelista, line 174:
```suggestion
Molecular systems with conjugated π bonds are often arranged into planar geometries.
```

docs/source/methods/avas.rst
```diff
@@ -168,6 +168,241 @@ We then use this guess of active orbitals to compute the CASCI energy:::
...
+For such systems, the active orbitals are usually chosen as the π orbitals perpendicular to the plane.
```
fevangelista, line 175:
```suggestion
For such systems, it often desirable to select an active space that includes π orbitals perpendicular to the plane.
```

docs/source/methods/avas.rst
```diff
@@ -168,6 +168,241 @@ We then use this guess of active orbitals to compute the CASCI energy:::
...
+Each π orbital consists of atomic p orbitals, which are also perpendicular to the plane.
```
fevangelista, line 176:
```suggestion
Each π orbital is a linear combination of atomic p orbitals, which are also perpendicular to the plane.
However, unlike the case of formaldehyde, where it easy to select the appropriate π and π* orbitals, in the more general case a π orbital is a linear combination of 2px, 2py, and 2px orbitals.

Fore example, in the following formaldehyde geometry, the atoms do not lie on the xy, xz, or yz plane ::
```

docs/source/methods/avas.rst
```diff
@@ -168,6 +168,241 @@ We then use this guess of active orbitals to compute the CASCI energy:::
...
+Take the above formaldehyde as an example.
+The π and π* orbitals are built from the 2p orbitals of C and O that are perpendicular to the molecular plane.
+We are very lucky that the formaldehyde molecule is in yz plane and
+the subspace 2p orbitals can be easily chosen as the 2px orbitals of C and O atoms.
+
+In reality, such perfect alignment does not happen often.
+Consider the below formaldehyde geometry ::
```
fevangelista, line 183:
```suggestion
```

docs/source/methods/avas.rst
```diff
@@ -168,6 +168,241 @@ We then use this guess of active orbitals to compute the CASCI energy:::
...
+Defining the molecular plane for π orbitals
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Molecular systems with conjugated π bonds generally process planar geometries.
+For such systems, the active orbitals are usually chosen as the π orbitals perpendicular to the plane.
+Each π orbital consists of atomic p orbitals, which are also perpendicular to the plane.
+Take the above formaldehyde as an example.
+The π and π* orbitals are built from the 2p orbitals of C and O that are perpendicular to the molecular plane.
+We are very lucky that the formaldehyde molecule is in yz plane and
+the subspace 2p orbitals can be easily chosen as the 2px orbitals of C and O atoms.
```
fevangelista, line 180:
```suggestion
```

docs/source/methods/avas.rst
```diff
@@ -168,6 +168,241 @@ We then use this guess of active orbitals to compute the CASCI energy:::
...
+    *   A      1    2      0.621379
+    *   A      8    0      0.374453
+    *   A      9    0      0.087267
+    ===============================
+
+Pay attention to the overlap metric.
+The projector becomes less effective and it is unclear which is the π orbital.
+The π* orbital also has a small overlap (0.374), much smaller than that from the previous aligned geometry (0.993).
+The reason for this is simply the subspace orbitals (i.e., 2px of C and O)
+are not perpendicular to the molecular plane.
```
fevangelista, line 213: I think this example does not necessarily help. Sure, it explains what happens when you do the wrong thing and invoke AVAS on a nonplanar geometry. But this is something that I would explain pro-actively, in the sense that if we want the user to avoid an error, we should provide a list of few important things to do before running AVAS. I would jump right directly to the next section and directly say "The approach described in the previous section is not flexible enough to treat general π systems like molecules containing multiple π systems or π systems that are not aligned with a specific molecular axis."

docs/source/methods/avas.rst
```diff
@@ -168,6 +168,241 @@ We then use this guess of active orbitals to compute the CASCI energy:::
...
+
+There are two ways to fix this problem.
+One is to reorient the molecule such that the molecular plane lying in yz plane.
+However, this approach is not flexible enough to treat multiple π systems in a molecule.
+The other approach is to use all px, py, pz orbitals as basis,
+using which the p orbital perpendicular to the plane can be defined.
+To do this, we need to specify two keywords: :code:`SUBSPACE` and :code:`SUBSPACE_PI_PLANES`.
+The latter is used to define the π plane, while we need all p orbitals for the former
```
fevangelista, line 221:
```suggestion
The latter is used to define the π plane, while we need all 2p orbitals for the former
```

fevangelista: I suggest adding some important details here.

docs/source/methods/avas.rst
```diff
@@ -168,6 +168,241 @@ We then use this guess of active orbitals to compute the CASCI energy:::
...
+For a more realistic example, consider the following iron porphyrin related molecule:
+
+.. image:: images/FeP.png
+    :width: 600
+    :align: center
+    :alt: An iron porphyrin complex.
+
+This molecule contains two π systems, namely, porphyrin and imidazole.
+Also, the porphyrin is not a perfect plane anymore.
+The following snippet selects all 3d orbitals of Fe, 3p orbitals of S,
```
fevangelista, line 295:
```suggestion
The following snippet selects all the 3d orbitals of Fe, the 3p orbitals of S,
```

docs/source/methods/avas.rst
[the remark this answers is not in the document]
fevangelista: Still, I recommend to add a line to say that ["Fe","C6-25","N3-6"] defines the porphyrin plane, while ["N1-2","C1-3"] defines the imidazole ring. Otherwise the reader has to do the mental work of figuring out what these lists mean.

docs/source/methods/avas.rst
```diff
@@ -168,6 +168,241 @@ We then use this guess of active orbitals to compute the CASCI energy:::
...
+    SOCC ACTIVE         0
+    UOCC ACTIVE        15
+    UOCC INACTIVE     462
+    ---------------------
+    RESTRICTED_DOCC   106
+    ACTIVE             37
+    RESTRICTED_UOCC   462
+    ---------------------
+
+  ==> Atomic Valence MOs (Active Marked by *) <==
```
fevangelista, line 325: Could we condense this table with lines that say "..."? Would the point we want to make be sacrificed if we shrink it?

docs/source/methods/avas.rst
```diff
@@ -168,6 +168,241 @@ We then use this guess of active orbitals to compute the CASCI energy:::
...
+        subspace_pi_planes [["C","O","H"]]
+        avas               true
+        avas_diagonalize   true
+        avas_cutoff        0.5
+      }
+
+    Here I prepare a basis called "double-shell.gbs", which includes the 2p and 3p orbitals of C and O atoms.
+    You can also prepare your own MINAO basis by truncating the the cc-pVTZ or ANO-RCC-VTZP basis sets.
+
+For a more realistic example, consider the following iron porphyrin related molecule:
```
fevangelista, line 286: I think this should be a new section specifically for molecules with two or more pi planes. You also need to explain (perhaps at the end, how AVAS uses this information). There are a lot of technical details at the beginning of the text, but here we do not explain what the multiple planes do or why they are necessary.

forte/orbital-helpers/aosubspace.h
```diff
@@ -127,58 +131,51 @@ class AOInfo {
...
+    ///   px, py, pz (3 orbitals) -> nx * px + ny * py + nz * pz (1 orbital),
+    ///   where the atom normal is a 3D unit vector (nx, ny, nz).
+    ///
+    ///   If the map is empty, the p orbitals are in the xyz frame defined the molecule.
+    ///   Equivalently, 3 normals are attached to each atom: (1,0,0), (0,1,0), and (0,0,1).
+    ///   As such, we still have a full set of p orbitals in the subspace.
     AOSubspace(std::vector<std::string> subspace_str, std::shared_ptr<psi::Molecule> molecule,
-               std::shared_ptr<psi::BasisSet> basis);
+               std::shared_ptr<psi::BasisSet> basis,
+               std::map<std::pair<int, int>, psi::Vector3> atom_normals = {},
```
fevangelista, line 160: The choice of passing the `pair<int,int>` implies some sort of processing inside the class, where one then has to find which atom maps to which atom in the molecule. I understand that this might take extra time, but a consistent solution is to pass a map<std::vector<std::string>,psi::Vector3> ( for example {["C","H","O"] : (1.0, 0.0,0.0)} ).
Is that hard to do?

forte/orbital-helpers/aosubspace.h
```diff
@@ -127,58 +131,51 @@ class AOInfo {
...
+    ///   This argument is used to make the subspace p orbitals of an atom aligned to the normal:
+    ///   px, py, pz (3 orbitals) -> nx * px + ny * py + nz * pz (1 orbital),
+    ///   where the atom normal is a 3D unit vector (nx, ny, nz).
+    ///
+    ///   If the map is empty, the p orbitals are in the xyz frame defined the molecule.
+    ///   Equivalently, 3 normals are attached to each atom: (1,0,0), (0,1,0), and (0,0,1).
+    ///   As such, we still have a full set of p orbitals in the subspace.
     AOSubspace(std::vector<std::string> subspace_str, std::shared_ptr<psi::Molecule> molecule,
-               std::shared_ptr<psi::BasisSet> basis);
+               std::shared_ptr<psi::BasisSet> basis,
```
fevangelista, line 159: I'd change the variable name to suggest that this is the MINAO
```suggestion
               std::shared_ptr<psi::BasisSet> minao_basis,
```

fevangelista: and similarly in the documentation.

forte/orbital-helpers/aosubspace.h
```diff
@@ -127,58 +131,51 @@ class AOInfo {
...
-    ///   2. Atom label, e.g. "C" (string)
-    ///   3. Atom count, e.g. 3 = third atom of a given kind (int)
-    ///   4. Energy level (n), 2 = 2s or 2p (int)
-    ///   5. l/m label, e.g. "2px" (string)
-    ///
-    /// @arg str_format A string that specifies the output formatting
-    std::vector<std::string> aolabels(std::string str_format = "%2$s%3$d (%4$d%5$s)") const;
-
-    /// Return a vector of AOInfo objects
-    const std::vector<AOInfo>& aoinfo() const;
```
fevangelista, line 173: Are all these delete functions obsolete or were they reproduced in the python side?

tests/methods/avas-7/input.dat
fevangelista: I think the comment suggests that it's not obvious what is going on here. A simple comment like "Selected the molecular plane defined by all the C, O, and H atoms" would do.

tests/methods/tests.yaml
fevangelista: This is a good point. I think that for the C20 example you should add more details. For example, it would be more clear to point out that here we select a different plane for each carbon atom, where each plane is defined by the nearest neighbors of each atom.

tests/methods/avas-8/input.dat
```diff
@@ -0,0 +1,112 @@
+# Test the pi orbital selection of AOSubspace on overlapped planes C20 fullerene
+
+import forte
```
fevangelista, line 3: Here you could add:
```
# An example of how AVAS can be used to select the π orbitals of the C20 fullerene.
# Here we use the nearest and second nearest neighbors of each atom to determine a plane normal to the C20 cage.
```

fevangelista:
@lcyyork : I left one more batch of suggestions. This is a great PR, thanks for adding this new functionality to Forte! I'd like to talk to you about how to automate this further in the future.

forte/orbital-helpers/aosubspace.h
lcyyork: I thought about this name. I did not use this name because 1) we used to use minimal/large basis for this. 2) if I want to select double d-shell or double pi active space, the basis must not be MINAO.

forte/orbital-helpers/aosubspace.h
lcyyork: I use `pair<int, int>` to **prevent** any processing of `atom_normals` inside the `AOSubspace` class, because `atom_to_aos_` is a vector of vector, i.e, atomic Z to a list of AO number. The processing is done outside the class.

The loop structure of `AOSubspace` is looping over the subspace orbitals, e.g., `{"C(2p)", "O(2p)"}`. For each entry (e.g., `"C(2p)"`), loop over the atom indices in the molecule. For each atom, loop over the AOs and push the matching AO to subspace.
I only need to add a few `if`s to make the atom normals to fit in this loop structure.

Although I am not a big fan of the current data structure, changing this would mean an overhaul to the code.

forte/orbital-helpers/aosubspace.h
lcyyork: They are not used at all, which brings down the codecov.

tests/methods/avas-8/input.dat
lcyyork: Changed.

docs/source/methods/avas.rst
lcyyork: I added a new section. Hope the change matches your standard.

tests/methods/avas-7/input.dat
lcyyork: I explained more in the text.

docs/source/methods/avas.rst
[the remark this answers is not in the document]
lcyyork: Text added afterwards.

forte/orbital-helpers/aosubspace.cc
lcyyork: I think `has_puream` is `false` if using `cartesian` in the basis but `true` for `spherical`. I do not like the ordering of `lm_labels_spherical_` because I have to consider the mismatch between it (pz, px, py) and the normal vector (nx, ny, nz).

fevangelista (approved):

PR #204, opened 2020-11-13
Internal Excitations for GAS-DSRG

lcyyork:
## Description
The major goal of this PR is to add internal amplitude support to GAS-DSRG-PT2/PT3. Other changes related to this goal happen in the `SemiCanonical` class. Address Future Work of PR #202, part of TODO list of PR #182 (Issue #185).

## Detailed Considerations
In the actual PT2/PT3 implementation (not the one within `MRDSRG` code), we assume the zeroth-order Hamiltonian (H0th) being the diagonal elements of Fock operator under semicanonical orbitals. This is equivalent to say that _H0th contains only the diagonal blocks (C-C, V-V, GAS1-GAS1, GAS2-GAS2, ..., GAS6-GAS6) of the Fock matrix_. As such, GASm-GASn (m != n) blocks of Fock become first order. (Note that all GAS orbitals are considered as **ACTIVE** in the DSRG.)

SInce there are restrictions of number of electrons in different GAS spaces, the number of determinants/configurations is greatly reduced than doing a CAS for the entire active orbitals. To address the electron correlation due to these missing determinants, one can possibly perform an unrestricted MRCI within the active space. Here, we do a crude approximation to that by considering **internal** excitations (GASm to GASn with _m_ < _n_) within the DSRG-MRPT framework, that is, we perform _a DSRG regularized internally contracted perturbation theory on top the GAS reference within the active space._ As a result, higher excitations (than those GASCI included) are considered within the active space.

The other option is to use another definition of semicanonicalization, where we no longer distinguish different GAS spaces but treat them together as active (just like in CAS). This will change the GASCI reference energy, but it is still a valid way to perform perturbation theory. To achieve this, we need to add an option to `Semicanonical` code for allowing mixing between GAS spaces.

Other options are added to `Semicanonical` class. **1)** the ability to treat `INACTIVE` orbitals (`FROZEN` + `RESTRICTED`) as a whole. **2)** we add the functionality of orbital rotations to natural orbitals (related to PR #200).

## User Notes
- [x] Add internal amplitudes in spin-adapted code.
- [ ] Add internal amplitudes in spin-integrated DSRG code.
[...]

PR #266, opened 2021-07-20
Natural orbital support for Semicanonical class

lcyyork:
## Description
Add natural orbital support for `Semicanonical` class. This is fraction of PR #204.

Another important feature added is to test CI overlap between two consecutive DSRG reference relaxation steps. If the targeted states altered, it will tries to restore the state by swapping the state weights. If the targeted cannot be located, the code will print a warning to the output file. (Sorry this is irrelevant to this PR but I do not want to wait and open another one due to the review of the paper.)

## User Notes
- [x] Possible to treat inactive orbitals (frozen + restricted) as a whole block
- [x] Possible to treat all GAS orbitals as a whole block
- [x] MCSCF prints warnings if suspicious rotations are detected.
- [x] Minor changes to related files due to CLion tidy check.
- [x] L-BFGS only add vectors if the dot product of difference vectors are positive
- [x] Several MCSCF default options are changed due to my experience.
- [x] Add single-particle basis transformation to RDMs class.
- [x] DSRG reference relaxation tries to follow states of the original CASCI.

## Checklist
- [x] Documented source code
- [x] Ready to go!


forte/base_classes/rdms.h
```diff
@@ -352,8 +358,8 @@ void make_cumulant_L3bbb_in_place(const ambit::Tensor& g1b, const ambit::Tensor&
  * @param Enuc the nucleaer repulsion energy
  * @return the reference energy
  */
-double compute_Eref_from_rdms(RDMs& ref, std::shared_ptr<ForteIntegrals> ints,
-                              std::shared_ptr<MOSpaceInfo> mo_space_info);
+double compute_Eref_from_rdms(RDMs& ref, const std::shared_ptr<ForteIntegrals>& ints,
```
fevangelista, line 361: I noticed that you use `const std::shared_ptr<T>` here. This just means that the pointer is constant. If you want to enforce that the integral object is constant (e.g. to signal that the orbitals should not change) you need to pass `std::shared_ptr<const T>`. I assume that this would require a type conversion from `const std::shared_ptr<T>` to `std::shared_ptr<const T>`, but I am not sure about how this works.

fevangelista: Ok, I found it: "shared_ptr<T> is convertible to shared_ptr<const T> but not the reverse."

lcyyork: This is just what CLion suggests for tidiness. I can change them for all at the end.

fevangelista: I see, does CLion give you an explanation for the suggestion? For what I see, it is suggesting to protect the shared pointer (which is OK). Probably the best thing to do would be `const std::shared_ptr<const ForteIntegrals>` but this starts to be verbose. :)

lcyyork: Actually is from clang-tidy. This is the message I get: `Clang-Tidy: The parameter 'ints' is copied for each invocation but only used as a const reference; consider making it a const reference`

fevangelista:
There are just two tests failing. When you fix them I'll be happy to review.

forte/base_classes/mo_space_info.h
```diff
@@ -188,6 +188,8 @@ class MOSpaceInfo {
     std::map<std::string, std::vector<std::string>> composite_space_names() const {
         return composite_spaces_;
     }
+    /// @return The names of nonzero GAS spaces
+    std::vector<std::string> nonzero_gas_spaces();
```
fevangelista, line 192: Please make `const`.

forte/base_classes/mo_space_info.h
```diff
@@ -202,6 +204,8 @@ class MOSpaceInfo {
     /// @return The list of the relative index (h,p_rel) of the molecular
     /// orbitals in space
     std::vector<std::pair<size_t, size_t>> relative_mo(const std::string& space) const;
+    /// @return True if the space is contained in a larger composite space
+    bool contained_in_space(const std::string& space, const std::string& composite_space);
```
fevangelista, line 208: Please make `const`

PR #267, opened 2021-07-28
Enable rotation between the orbitals in CASSCF_ACTIVE_FROZEN_ORBITAL

huangm29:
## Description
This code now enables rotations between orbitals within the list CASSCF_ACTIVE_FROZEN_ORBITALS instead of completely freezing rotations of all the orbitals within the list.

## User Notes
- Added test sa-gasscf-4 to check this implementation is correct.
## Checklist
- [x] Documented source code

### DIFF
--- a/forte/casscf/casscf_orb_grad.cc
+++ b/forte/casscf/casscf_orb_grad.cc
@@ -232,6 +232,16 @@ void CASSCF_ORB_GRAD::read_options() {
                 zero_rots_[irrep][nu].emplace(p);
                 zero_rots_[irrep][p].emplace(nu);
             }
+
+            // enable rotation between the orbitals within CASSCF_ACTIVE_FROZEN_ORBITAL
+            for (size_t j = 0; j < i; ++j) {
+                size_t v = frza_rot[j];
+                if (actv_rel_mos[v].first == irrep) {
+                    auto ki = actv_rel_mos[v].second;
+                    zero_rots_[irrep][nu].erase(ki);
+                    zero_rots_[irrep][ki].erase(nu);
+                }
+            }
         }
     }
 
--- a/forte/orbital-helpers/orbitaloptimizer.cc
+++ b/forte/orbital-helpers/orbitaloptimizer.cc
@@ -366,14 +366,16 @@ void OrbitalOptimizer::orbital_gradient() {
         std::set_difference(active.begin(), active.end(), active_frozen.begin(),
                             active_frozen.end(), std::inserter(unfrozen, unfrozen.begin()));
 
-        for (int i : active_frozen) {
-            for (int j : active_frozen) {
-                size_t ii = active_abs_[i];
-                size_t jj = active_abs_[j];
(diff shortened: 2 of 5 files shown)


forte/orbital-helpers/orbitaloptimizer.cc
```diff
@@ -366,14 +366,16 @@ void OrbitalOptimizer::orbital_gradient() {
...
 
-        for (int i : active_frozen) {
-            for (int j : active_frozen) {
-                size_t ii = active_abs_[i];
-                size_t jj = active_abs_[j];
-                Orb_grad_Fock->set(nhole_map_[ii], npart_map_[jj], 0.0);
-                Orb_grad_Fock->set(nhole_map_[jj], npart_map_[ii], 0.0);
-            }
-        }
+        // This part is removed to enable the rotation between the orbitals within
```
fevangelista, line 369: Can you please remove the code commented here?

fevangelista (changes requested):
The PR looks good. I have one question: are `nhole_map_` and `npart_map_` obsolete or are they still needed?

forte/casscf/casscf_orb_grad.cc
```diff
@@ -232,6 +232,16 @@ void CASSCF_ORB_GRAD::read_options() {
                 zero_rots_[irrep][nu].emplace(p);
                 zero_rots_[irrep][p].emplace(nu);
             }
+
+            // enable rotation between the orbitals within CASSCF_ACTIVE_FROZEN_ORBITAL
+            for (size_t j = 0; j < i; ++j) {
```
lcyyork, line 237: Should we add an option for this type of behavior? I am OK to take this as the default but we need to document this in the MCSCF documentation.

forte/orbital-helpers/orbitaloptimizer.cc
huangm29: Removed. The two `nhole_map_` and `npart_map_` variables are still needed.

huangm29:
This option is a little bit tricky because in general, you would put the entire GAS space which contains all core electrons into CASSCF_ACTIVE_FROZEN_ORBITAL. In this case, the rotation within a single GAS is not making any difference. The only case where this option can make a difference is when you put core electrons (where you want to freeze the rotation) into multiple GAS. 

The example I am thinking of here is the C K-edge for a pyridine molecule where there are three different types of C atoms, and I would like a specific transition from one type of carbon core orbital. For example, I use one C 1s as GAS1 (the one is opposite to N), and the other 4 as GAS2, and frozen GAS1 occupation to 1. Should I freeze the rotation between GAS1 and GAS2? My original thought was I should NOT freeze them, but now I think this is dangerous so as it can still lead to a variational collapse. In other words, I think this PR may better be closed.

BTW, the current way I am running pyridine calculation is to put all C 1s orbital in a single GAS and do state-average calculations. This calculation I mentioned might just be an alternative approximation tool for assignment, double-checking, etc.

Issue #268, opened 2021-08-03
Computing RDMs Segfault for FCI Class


lcyyork:
The following input should reproduce the error.
```
molecule {
2 5
Fe
symmetry c1
}

set globals{
  basis                  cc-pvdz
  reference              rohf
  scf_type               df
  e_convergence          10
  d_convergence          5
  ints_tolerance         0.0
}
Escf, wfn = energy('scf', return_wfn=True)

set forte{
  int_type               df
  active_space_solver    fci
  correlation_solver     sa-mrdsrg
  frozen_docc            [5]
  restricted_docc        [4]
  active                 [5]
  e_convergence          12
  r_convergence          8
}
energy('forte', ref_wfn=wfn)
```

The active space solver would create three `FCI` objects, one for each spin Ms that is greater than 0.
We will compute the RDMs for each Ms value and average them, which generates a segfault.
The issue is that for Ms = 0, the max size of string list is 10; but for Ms = 2, the max size becomes 5.
The energy computation starts from Ms = 0 to Ms = 1 to Ms = 2.
When computing the RDMs in `fci_vector_rdm.cc`, a temporary matrix `C` is set of `C1`, for example, line 162.
However, `C1` is marked as `static` in line 168 of `fci_vector.h`, which is a 5 x 5 matrix after the Ms = 2 energy computation.
Now if I want to compute the RDMs for Ms = 0, here goes the segfault.

@fevangelista  So I wonder if we should make the change to `FCI` class or not.
For my purpose (computing Ms-averaged RDMs), I can leave FCI as is and implement S+ or S- operator for the spin-complete active methods.

fevangelista:
This should definitively be fixed. Since you have identified the main issue, it should be easy for me to solve it.

lcyyork:
I tried to simply modify the dimension, but it got stuck for Ms = 2. I did not go further from there.

rwashi690:
Yeah, I ran into this issue and as well with MnH septet state - here is my input and output files

```
import forte
molecule dimer{
0  7
Mn
H  1 R
}
dimer.R = 1.7309
set globals{
basis               cc-pVQZ
reference           rohf
scf_type            df
DOCC                [6, 0, 2, 2]
SOCC                [3, 1, 1, 1]
maxiter             250
d_convergence       1e-2
e_convergence       1e-2
damping_percentage  20
}
Escf, scf_wfn = energy('scf', return_wfn=True)
set forte {
job_type            none
avas                true
[...]

PR #269, opened 2021-08-04
Fix a bug in FCIVector memory handling

fevangelista:
This PR fixes issue #268.

## Description
Now the `FCIVector` avoids reducing the size of two temporary matrices and avoid memory allocation if not needed.

## User Notes
- [x] Fix issue #268 

## Checklist
- [x] Ready to go!

### DIFF
--- a/forte/fci/fci_vector.cc
+++ b/forte/fci/fci_vector.cc
@@ -55,8 +55,12 @@ double FCIVector::h2_bbbb_timer = 0.0;
 
 void FCIVector::allocate_temp_space(std::shared_ptr<StringLists> lists_, int print_) {
     // TODO Avoid allocating and deallocating these temp
-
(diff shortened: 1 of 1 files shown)


lcyyork (commented):
It might be a good idea to add a test case.

tests/methods/fci-10/input.dat
```diff
@@ -0,0 +1,34 @@
...
+  d_convergence          5
+  ints_tolerance         0.0
+}
+Escf, wfn = energy('scf', return_wfn=True)
+
+set forte{
+  int_type               df
+  active_space_solver    fci
+  spin_avg_density       true
+  #correlation_solver     sa-mrdsrg
```
lcyyork, line 25: I think the RDMs have to be computed to detect this bug. The current test actually passes without the fix from this PR.

PR #270, opened 2021-08-10
Adjoint for SQOperator and SparseOperator

JonathonMisiewicz:
## Description
This PR adds an adjoint method to the `SQOperator` and `SparseOperator` classes, so that I can use the SparseOperator API to correctness-test truncated VCC.

## User Notes
- [x] `SQOperator` and `SparseOperator` now have a defined `adjoint` method

## Checklist
- [x] Added pytests, so no output files
- [x] Ready to go!


fevangelista (approved):
Thanks for adding this function!

PR #274, opened 2021-11-15
Update Forte to work with development Psi4

JonathonMisiewicz:
## Description
Updates Forte to work with newer version of Psi4. Some recent Psi PRs broke Forte due to:
* Switching to scoped namespaces for DIIS enums, which is the C++ standard
* Switching `Vector::clone` from returning raw pointers to unique pointers. This exposed a memory leak in Forte, as well as very clumsy code. Both are fixed.

The new code compiles, and if auto-tests pass, this should be good to go.

## Checklist
- [x] Ready to go!

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -137,9 +137,6 @@ jobs:
       cd build
       echo "Now at directory: $PWD"
       git clone https://github.com/psi4/psi4.git psi4
-      cd psi4
-      git checkout de731dbf19932092603c25a1e2a02cf813ab77ee
-      cd ..
     displayName: 'Clone Psi4 Source'
 
   - bash: |
--- a/forte/casscf/casscf.cc
+++ b/forte/casscf/casscf.cc
@@ -186,9 +186,9 @@ double CASSCF::compute_energy() {
 
     // Setup the DIIS manager
     auto diis_manager = std::make_shared<DIISManager>(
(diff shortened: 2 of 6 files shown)


fevangelista:
@JonathonMisiewicz: I pushed some more changes to make sure we save the updated MOs in ForteIntegrals. When you have a chance can you please remove the following from the azure-pipelines.yaml file to enable testing on the latest psi4 commit?
```
      cd psi4
      git checkout de731dbf19932092603c25a1e2a02cf813ab77ee
      cd ..
```

fevangelista (approved):
Once the code to checkout an older commit of psi4 is removed this should be ready to go.

JonathonMisiewicz:
Thoughts on the test failure? I can't imagine why Py37 or Py39 would make a difference, or how updating Psi could cause that.

fevangelista:
It's a failure of psi4. Here is the incriminating part:
```
  ==> Starting MCSCF iterations <==

        Iter         Total Energy       Delta E   Orb RMS    CI RMS  NCI NORB
   @MCSCF  1:                 nan    nan  nan  0.00e+00   12    1  Initial CI
   @MCSCF  2:                 nan    nan  nan  0.00e+00   12    1  TS
   @MCSCF  3:                 nan    nan  nan  0.00e+00   12    1  TS
   @MCSCF  4:                 nan    nan  nan  0.00e+00   12    1  TS, DIIS
```

I am going to merge this. Our solutions should probably be to switch to using forte's CASSCF code instead of psi4's.

PR #278, opened 2021-12-14
DF-MP2 natural orbitals

lcyyork:
## Description
DF-MP2 natural orbitals

## User Notes
- [x] Add DF support for MP2 natural orbitals
- [x] Add a function to split the vector to several subvectors
- [x] Other minor changes or typo fixes

## Checklist
- [x] Documented source code
- [x] Ready to review 
- [x] Ready to go!


JonathonMisiewicz:
Why are we building the MP2 natural orbitals in Forte rather than getting the natural orbitals from Psi?

lcyyork:
> Why are we building the MP2 natural orbitals in Forte rather than getting the natural orbitals from Psi?

I am not sure how to run MP2 natural orbital using Psi4. Do you have an input example?

lcyyork:
Thank you for the suggestion. By SO, do you mean symmetry-adapted AO?

Another to clarify is that Forte does support MP2-NO, just without DF. I just want it to suggest an active space for me.

JonathonMisiewicz:
_sigh_ Good question. Neither code supports symmetry, which is another obstacle to Forte using them for active space purposes. For DF-MP2 with symmetry, you would need a new code.

I meant "Symmetry Orbital" when I wrote my last message, but "Atomic Orbital" would be more accurate.

fevangelista:
Nice contribution. I think that in the long run we need to refactor this, maybe adding MP2 to Forte as its own method. Will put on my list of things to review.

fevangelista:
The MP2 NOs test case seems to be failing. I'll wait for you to fix it before taking a look.

lcyyork:
> The MP2 NOs test case seems to be failing. I'll wait for you to fix it before taking a look.

I think it's fixed. Please take a look when you have some time to kill.

fevangelista (approved):

PR #286, opened 2022-02-11
Changes for Forte to compile on almost current master Psi

JonathonMisiewicz:
## Description
Per @fevangelista request, this PR allows Forte to compile with master Psi, almost. That also requires a Psi PR.

I'm quite pleased with the code cleanup.

## User Notes
- [x] Forte is compatible with current master Psi, but not older versions

## Checklist
- [x] Ready to go!


fevangelista:
Can you please try to remove the lines:
```
      cd psi4
      git checkout 2feab68a2ce052ed74b75fa6749090cc467cd1d9
      cd ..
```
from `azure-pipelines.yml`?

JonathonMisiewicz:
I tried, but the attempt was not successful. The issue seems to have something to do with Pybind. I can poke further at this if you want. The cause of this isn't obvious to me.

fevangelista:
Thank you. Now the error is different. Error messages like `error: ‘basic_string_view’ is not a member of ‘std’` suggest that there is a problem with setting the C++ standard. basic_string_view is part of C++17. I could not spot any issue in the azure file. Maybe something changed in psi4 with l2?

JonathonMisiewicz:
That fixed build failures. Test failures will be fixed by [a Psi PR](https://github.com/psi4/psi4/pull/2441/files).

fevangelista:
Thanks for working on this PR!

JonathonMisiewicz:
Unfortunately, the PR wasn't finished, and there will need to be a follow-up to get the remaining tests passing.

I was going to work on this today but hit a wall when Forte wouldn't even run for me locally.

PR #289, opened 2022-03-10
fix rdms when doing dipole

lcyyork:
## Description
Fixed incorrect CASCI dipole after DSRG due to inconsistent orbital basis.

## Checklist
- [x] Documented source code
- [x] Ready to go!


forte/base_classes/active_space_solver.h
```diff
@@ -199,6 +205,10 @@ class ActiveSpaceSolver {
     /// Read wave function from disk as initial guess
     bool read_initial_guess_;
 
+    /// Unitary matrix for orbital rotations
```
fevangelista, line 208: Please make it completely clear what's the role of these matrices and where they are used. For example (and I am not sure this is even correct)
```suggestion
    /// Unitary matrices that account for orbital rotations due to semi-canonicalization
    /// These are used in ... but only for ...
```

fevangelista (changes requested):
Thanks for this PR. I rebased to master and it seems to run. Is there any test that checks the values? If not, can you add one to this PR?

lcyyork:
Hi Francesco, this PR is not ready. I need to fix permanent dipole as well. The issue is the orbital mismatch between ActiveSpaceIntegrals and ForteIntegrals. I am thinking if the semi-canonicalization should only happen inside DSRG (or wherever needed) without touching ForteIntegrals.

tests/methods/gasci-dsrg-mrpt2-2/input.dat
```diff
@@ -0,0 +1,52 @@
...
+  gas2                  [5,0,2,2]
+  gas1max               [2,1,1]
+  gas1min               [2,1,1]
+  transition_dipoles    true
+  e_convergence         8
+  r_convergence         7
+}
+edsrg = energy('forte', ref_wfn = wfn)
+
+#psi4.core.print_variables()
```
fevangelista, line 40:
```suggestion
```

tests/methods/gasci-dsrg-mrpt2-2/input.dat
```diff
@@ -0,0 +1,52 @@
...
+
+dip_a1 = variable("SINGLET <0A1|DM_Z|0A1> ENTRY 0")
+dip_b1 = variable("SINGLET <0B1|DM_Z|0B1> ENTRY 0")
+dip_b2 = variable("SINGLET <0B2|DM_Z|0B2> ENTRY 0")
+compare_values(dip_a1, variable("SINGLET <0A1|DM_Z|0A1> ENTRY 1"), 8, "Dipole Z singlet A1 root 0")
```
fevangelista, line 45: This does not look like a comparison with a reference number. Can you please compare to a numerical value instead so that we can detect issues with the code in the future?

fevangelista (commented):
I just a few small things that should be taken care of before committing this PR. When commuting the PR please squash and merge.

tests/methods/gasci-dsrg-mrpt2-2/input.dat
lcyyork: This is actually the intended test that the dipole moments should not be changed before and after DSRG with s = 0 due to orbital changes.

fevangelista: Ok, but you can imagine cases under which this test will pass but the dipole is wrong for other reasons. I would then add a comment to the test to explain what you are testing.

Issue #288, opened 2022-02-16
Forte Compilation Error due to DF RMP2


JonathonMisiewicz:
Forte doesn't compile for me, even after following the [readme](https://github.com/evangelistalab/forte/blob/master/README.md) instructions for `setup.py`. I compile successfully upon commenting out the declarations and definitions and calls to
```
    /// Build OO part of DF-UMP2 1-RDM
    void compute_df_ump2_1rdm_oo(ambit::BlockedTensor& D1);
    /// Build VV part of DF-UMP2 1-RDM
    void compute_df_ump2_1rdm_vv(ambit::BlockedTensor& D1);
    /// Build OO part of DF-RMP2 1-RDM
    void compute_df_rmp2_1rdm_oo(ambit::BlockedTensor& D1);
    /// Build VV part of DF-RMP2 1-RDM
    void compute_df_rmp2_1rdm_vv(ambit::BlockedTensor& D1);
```

The offending functions may be found in `mp2_nos.h` and `mp2_nos.cc`. Error message is
```
ImportError: dlopen(/Users/jonathonmisiewicz/forte/forte/_forte.cpython-37m-darwin.so, 2): Symbol not found: __ZNSt3__113__invokable_rIvRZN5forte7MP2_NOS23compute_df_rmp2_1rdm_ooERN5ambit13BlockedTensorEE4$_15JRKNS_6vectorImNS_9allocatorImEEEERdEE5valueE
  Referenced from: /Users/jonathonmisiewicz/forte/forte/_forte.cpython-37m-darwin.so
  Expected in: flat namespace
 in /Users/jonathonmisiewicz/forte/forte/_forte.cpython-37m-darwin.so
```

Compilation fail happens even on pre-DIIS refactor and L2 Psi.

JonathonMisiewicz:
Including [an iterate statement like this](https://github.com/evangelistalab/forte/blob/master/forte/orbital-helpers/mp2_nos.cc#L499-L501) is sufficient to make the crash occur, even when the actual contents of the function signature are commented out.

JonathonMisiewicz:
Removing `default(none) shared(Jab)` will also cure the import problem, though I'm puzzled as to _why_.

JonathonMisiewicz:
[...]

PR #290, opened 2022-03-26
Fix Forte [Forte Part]

JonathonMisiewicz:
## Description
This PR makes changes so that `Psi` can do DIIS on an `ambit.BlockedTensor`, as required by the `forte` plugin. This PR will not work until [this ambit PR](https://github.com/jturney/ambit/pull/53) and [this Psi PR](https://github.com/psi4/psi4/pull/2505) are merged, but passed locally on a _slightly_ outdated version of Forte. 

* As a temporary fix to #288, this PR replaces the OpenMP commands that cause runtime problems for me, locally. I'm happy to investigate #288 more with York, once this PR is on.
* The meaning of "UNRELAXED DIPOLE" changed from _dipole magnitude_ to _dipole matrix_. The corresponding "UNRELAXES DIPOLE X" and friends variables are removed. This synchronized dipole variables with their meanings in Psi.
* We now apply DIIS to `ambit.BlockedTensor` directly, with much cleaner code.
* Forte is no longer responsible for exposing `ambit.BlockedTensor` to Python. We now import whatever `ambit` exposes for us.


lcyyork (approved):
I am OK with the PR, thanks!

forte/api/forte_python_module.cc
```diff
@@ -70,7 +70,6 @@ using namespace pybind11::literals;
 namespace forte {
 
 // see the files in src/api for the implementation of the following methods
-void export_ambit(py::module& m);
```
fevangelista, line 73: It's great to see code removed!

setup.py
```diff
@@ -98,6 +98,7 @@ def build_extension(self, ext):
         cmake_args += [f'-DMAX_DET_ORB={self.max_det_orb}']
         cmake_args += [f'-DENABLE_CODECOV={str(self.enable_codecov).upper()}']
         cmake_args += [f'-DENABLE_ForteTests=TRUE']
+        cmake_args += ['-DCMAKE_OSX_SYSROOT=' + os.environ["CONDA_BUILD_SYSROOT"]]
```
fevangelista, line 101: This assumes the user is using conda?

tests/pytest/ambit/test_ambit.py
```diff
@@ -1,15 +0,0 @@
-#!/usr/bin/env python
-# -*- coding: utf-8 -*-
-
-import forte
-
-def test_ambit():
-    """Test the conversion of ambit::Tensor to numpy"""
-    t = forte.test_ambit_3d()
```
fevangelista, line 8: There must be a function called `test_ambit_3d()`. It looks like it will be no longer needed. Can you please remove it?

fevangelista (changes requested):
Thanks for all these changes. With the new DIIS code things are much more compact. I left a few comments. Let's first merge the ambit and psi4 changes and then merge this in forte.

setup.py
fevangelista: I wonder if we should make this optional.

JonathonMisiewicz: Good call - I need this line locally to build Psi and Forte. I'll make this into a comment instead.

tests/pytest/ambit/test_ambit.py
JonathonMisiewicz: [Already done.](https://github.com/evangelistalab/forte/pull/290/files#diff-8a41d74eddc62044d7b9cd2b0f5817cb68719d19f9acdc4ae5b7b3ed0e6ed90eL68-L82)

JonathonMisiewicz:
Ambit and Forte PRs are in, so tests should pass.

fevangelista (approved):
Wow, thanks for all this hard work! Happy to see the green checkmark back 😄.

PR #294, opened 2022-05-23
Adding Natural Transition Orbital analysis

huangm29:
## Description
This pr adds Natural Transition Orbitals(NTO) analysis for dipole moment. It also fixes the output of forte to make it better for core-excite calculations

## User Notes
- [x] Calculate the NTO rotation matrix which helps assigns a certain transition.
- [x] Adding option "PRINT_DIFFERENT_GAS_ONLY" where forte will only calculate transition intensities between states which have different occupations among different GAS.
- [x] Adding option "DUMP_TRANSITION_RDM" which allows printing of the transition reduced density matrix into a file
- [x] Change state_info so the ground electronic state is always before the core-excited states in the order of the roots. 
- [x] **(This feature has been changed so it does not require newest PR of psi4)** This PR requires the new psi4 patch [https://github.com/psi4/psi4/pull/2585#issue-1244059821](url) or add a line as"PSI_API" before function `int C_DGESDD` to enable the psi4 wrapper for SVD.
- [x] Adding tests for gasci-3 and gasci-dsrg-mrpt2-2 to check the maximum S_value from SVD of the transition reduced density matrix.

## Checklist
- [x] Documented source code
- [x] Ready to go!


forte/base_classes/active_space_method.cc
```diff
@@ -356,6 +358,71 @@ std::vector<std::vector<double>> ActiveSpaceMethod::compute_transition_dipole_sa
     }
     psi::outfile->Printf("\n    %s", dash.c_str());
 
+    // Doing SVD for transition reduced density matrix
+
+    // Several necessary matrix and constants for LAPACK SVD
```
lcyyork, line 363: Why not using libmint's svd function here?

fevangelista:
I agree with @lcyyork's comments. Please move to use the psi::Matrix class svd and see if you can add a test case. If you need help with that we can give you ideas.

fevangelista:
Also, it looks like there are issues with psi4-1.6/forte. Maybe we should try to fix it first before merging this PR.

forte/base_classes/active_space_method.cc
huangm29: I changed to libmint svd.

huangm29:
Test cases added.

huangm29:
I think this PR is ready now. Let us try to move this out of the way.

forte/base_classes/state_info.cc
```diff
@@ -69,9 +69,16 @@ const std::vector<size_t>& StateInfo::gas_min() const { return gas_min_; }
...
-    return std::tie(na_, nb_, multiplicity_, twice_ms_, irrep_, gas_min_, gas_max_) <
-           std::tie(rhs.na_, rhs.nb_, rhs.multiplicity_, rhs.twice_ms_, rhs.irrep_, rhs.gas_min_,
-                    rhs.gas_max_);
+    // Make sure the roots are in correct energy order
+    if ((gas_min_ == rhs.gas_min_) && (gas_max_ == rhs.gas_max_)) {
+        return std::tie(na_, nb_, multiplicity_, twice_ms_, irrep_, gas_min_, gas_max_) <
+               std::tie(rhs.na_, rhs.nb_, rhs.multiplicity_, rhs.twice_ms_, rhs.irrep_,
+                        rhs.gas_min_, rhs.gas_max_);
+    } else if (gas_max_ == rhs.gas_max_) {
+        return gas_min_ > rhs.gas_min_;
```
lcyyork, line 78: Seems to me this should be
```suggestion
        return gas_min_ < rhs.gas_min_;
```
?

lcyyork (approved):
I have a question about the comparison of `StateInfo`, which I could be wrong. Other than that, please clang format the code. Everything else looks fine to me.

forte/base_classes/state_info.cc
lcyyork: Thanks for the clarification. Might worth writing some comments here because this is not we used to have (the tuple comparison actually does `<`).

lcyyork:
@huangm29 Sorry Meng, I thought this has been merged. I wonder could you merge the latest forte and see if all tests pass. If so, please go ahead and squash and merge.

PR #295, opened 2022-05-28
Fix compatibility issues with Psi4 1.6

fevangelista:
## Description
Fix compatibility issues with Psi4 1.6

## User Notes
- [x] Changed azure configuration to compile ECPs
- [x] Disabled test cases that use psi4's findiff (temporarily broken due to the transition to the DD)
- [x] Fix one test case where the Ms value was not compute correctly (switch to `lround`)
- [x] Fix the `avas-6` test case where the initial SCF energy is slightly different (due to psi4 1.6?)
- [x] Fix error in ambit call that used self-assignment (needs a fresh pull from ambit)
- [x] Fix test case `fci-ex-1`

## Checklist
- [ ] Ready to go!

### DIFF
--- a/forte/base_classes/state_info.cc
+++ b/forte/base_classes/state_info.cc
@@ -114,11 +114,12 @@ StateInfo make_state_info_from_psi(std::shared_ptr<ForteOptions> options) {
     //    triplet: multiplicity = 3 -> twice_ms = 0 (ms = 0)
     size_t twice_ms = (multiplicity + 1) % 2;
     if (not options->is_none("MS")) {
-        twice_ms = std::round(2.0 * options->get_double("MS"));
+        twice_ms = std::lround(2.0 * options->get_double("MS"));
     }
 
-    if (((nel - twice_ms) % 2) != 0)
-        throw psi::PSIEXCEPTION("\n\n  make_state_info_from_psi: Wrong value of M_s.\n\n");
+    if (((nel - twice_ms) % 2) != 0){
+        throw std::runtime_error("\n\n  make_state_info_from_psi: Wrong value of M_s.\n\n");
+    }
 
     size_t na = (nel + twice_ms) / 2;
     size_t nb = nel - na;
--- a/forte/proc/dsrg.py
+++ b/forte/proc/dsrg.py
@@ -204,9 +204,12 @@ class ProcedureDSRG:
 
         # Reference relaxation procedure
         for n in range(self.relax_maxiter):
-            # Grab effective Hamiltonian in the active space
-            # These active integrals are in the original basis (before semi-canonicalize in the init function),
-            # so that the CI coefficients are comparable before and after DSRG dressing.
+            # Grab the effective Hamiltonian in the active space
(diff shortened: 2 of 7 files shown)


forte/base_classes/state_info.cc
```diff
@@ -114,11 +114,12 @@ StateInfo make_state_info_from_psi(std::shared_ptr<ForteOptions> options) {
     //    triplet: multiplicity = 3 -> twice_ms = 0 (ms = 0)
     size_t twice_ms = (multiplicity + 1) % 2;
     if (not options->is_none("MS")) {
-        twice_ms = std::round(2.0 * options->get_double("MS"));
+        twice_ms = std::lround(2.0 * options->get_double("MS"));
```
fevangelista, line 117: This was causing an error on my computer. Going to `lround` seems the right thing to do anyway.

forte/proc/dsrg.py
```diff
@@ -254,26 +256,34 @@ def compute_energy(self):
...
+                # Do NOT read previous orbitals if fixing orbital ordering and phases failed
                 if (not self.semi.fix_orbital_success()) and self.Heff_implemented:
                     psi4.core.print_out("\n  DSRG checkpoint files removed due to the unsuccessful"
                                         " attempt to fix orbital phase and order.")
                     self.dsrg_solver.clean_checkpoints()
-            self.Ua["ik"] = self.Ua["ij"] * self.semi.Ua_t()["jk"]
-            self.Ub["ik"] = self.Ub["ij"] * self.semi.Ub_t()["jk"]
 
-            # - Compute DSRG energy
+                # update the orbital transformation matrix that connects the original orbitals
```
fevangelista, line 278: Question for @lcyyork: indenting the lines below seems to be the right thing to do (apply Ua from semi to the current set of Ua only if we do indeed semicanonicalize). Otherwise we would be always rotating by some quantity (I assume the identity, but it's not good to introduce some indirect dependency on the internals of the semi class).

lcyyork: Yes, it is identity matrix by default.

lcyyork: > Question for @lcyyork: indenting the lines below seems to be the right thing to do (apply Ua from semi to the current set of Ua only if we do indeed semicanonicalize). Otherwise we would be always rotating by some quantity (I assume the identity, but it's not good to introduce some indirect dependency on the internals of the semi class).

Yes, it is identity matrix by default. The semi class is initialized no matter actual semi computations or not.

forte/proc/dsrg.py
```diff
@@ -254,26 +256,34 @@ def compute_energy(self):
...
+                # to the current semi-canonical ones. We do this only if we did a semi-canonicalization
+                temp = self.Ua.clone()
+                temp["ik"] = self.Ua["ij"] * self.semi.Ua_t()["jk"]
```
lcyyork, line 281: This could be `self.Ua["ik"] = temp["ij"] * self.semi.Ua_t()["jk"]`

fevangelista: Good point.

PR #304, opened 2022-10-28
Clean up DSRG transformed multipole integrals

lcyyork:
## Description
This PR adds DSRG transformed dipole and quadrupole integrals to spin-adapted MRDSRG. The computation is activated by the options `DSRG_MAX_DIPOLE_LEVEL` and `DSRG_MAX_QUADRUPOLE_LEVEL`, which are the max body considered in the recursive linear commutator approximation. Both 1 and 2 are supported for MRPT2 and LDSRG(2), while only 1 is supported for MRPT3.

For LDSRG(2), the formula is `Mbar = M + [M, A]_n + 0.5 * [[M, A]_n, A]_n + ...` where `n` is the max level kept for each commutator.

For PT2 and PT3, the formulae are different from those in the SA-DSRG-PT paper. In particular, I want the scalar of `Mbar` to be equal to the contraction between bare multipole integrals and the unrelaxed 1RDM. Define the unrelaxed 1RDM as the derivative of energy with respect to Fock matrix element, suggesting that the diagonal blocks of one-body integrals will be zeroth order (`Md`) while the off-diagonal blocks are of first order.

For PT2, `Mbar = M + [M, A1st]_n + 0.5 * [[Md, A1st]_n, A1st]_n`.

For PT3, `Mbar = M + [M, A1st]_n + [M, A2nd]_n + 0.5 * [[M, A1st]_n, A1st]_n + 0.5 * [[Md, A1st]_n, A2nd]_n + 0.5 * [[Md, A2nd]_n, A1st]_n + 1/6 * [[[Md, A1st]_n, A1st]_n, A1st]_n`.

The difference due to formulae change is more-or-less negligible for PT2. Take octatetraene as an example. The oscillator strengths of the singlet Bu states are
```
                     PT2                          PT3
      ------------------------------    ----------------------
       paper   max=2   max=1    bare     paper   max=1    bare
1Bu+  1.7418  1.7427  1.7114  1.8267    1.7888  1.7587  1.8873
1Bu-  0.0015  0.0015  0.0014  0.0017    0.0020  0.0019  0.0022
2Bu-  0.0017  0.0016  0.0016  0.0019    0.0004  0.0004  0.0004
```
where paper refers to the values of Table 1 in the 2018 SA-DSRG-PT2/PT3 paper, which were obtained by keeping the 2-body terms of `Mbar`. The above example shows that the difference due to formula change is at the third decimal while using max_level = 1 leads to smaller oscillator strengths. In either case, the results seem to be improved over those of using bare dipole integrals: 1.8267 and 1.8873 for PT2 and PT3, respectively.

The added test cases (`mrdsrg-spin-adapted-7` and `mrdsrg-spin-adapted-pt2-7`) compute the valence-Rydberg mixings of ethene reported in MS-CASPT2 paper. Here is the summary (excitation energies in eV and oscillator strengths in parentheses).
```
          CASSCF         PT2(bare/max=1/max=2)       PT3(bare/max=1)     LDSRG(2)(bare/max=1/max=2)
       ------------  ---------------------------  --------------------  ---------------------------
1B1u   8.07(0.3107)   8.13(0.4677/0.4165/0.4235)   7.93(0.4536/0.3567)   7.96(0.4572/0.3653/0.3819)
2B1u  11.17(0.2339)  11.16(0.0878/0.0715/0.0760)  11.03(0.0743/0.0440)  11.04(0.0764/0.0480/0.0523)

          EOM-CC2       EOM-CCSD    EOM-CC3
      -------------  -------------  -------
1B1u   7.88(0.4085)   8.00(0.3729)     7.90
2B1u  10.25(0.0903)  10.42(0.0832)    10.40
```
_LDSRG(2) data used NIVO approximation. The flow parameters are 0.35, 2.0, and 1.5 for PT2, PT3, and LDSRG(2), respectively._

## User Notes
- [x] Add DSRG transformed multipole integrals for spin-adapted MRDSRG.
- [x] Add `MultipoleIntegrals` and `ActiveMultipoleIntegrals` in `one_body_integrals.h/.cc`.
- [x] Optimize spin-adapted DSRG-MRPT2 energy.
- [x] Implement part of the unrelaxed 1RDM of DSRG-MRPT2 with large T2.
- [x] Add "pqQ" order for accessing a chunk of three-index integrals.
- [x] Implement in-place matrix transpose based on NVIDIA paper.
- [x] Add `canonicalize_B` and `read_Bcanonical` functions that transform B into semicanonical basis.

## Checklist
- [x] Documented source code
- [x] Ready to go!


fevangelista:
@lcyyork : what's the status of this PR?

lcyyork:
> @lcyyork : what's the status of this PR?

I am still working on it. There is ~3.0e-5 difference in DSRG dipole due to orbital rotation that I am trying to figure out now.

lcyyork:
@fevangelista This PR is almost ready. I will upload some test cases and modify the manual. I will also rerun all tests on my computer. Sorry for the large PR.

forte/base_classes/active_space_solver.h
```diff
@@ -132,6 +138,11 @@ class ActiveSpaceSolver {
         as_ints_ = as_ints;
     }
 
+    /// Pass multipole integrals to the solver (e.g. correlation dressed dipole/quadrupole)
+    void set_active_multipole_integrals(std::shared_ptr<ActiveMultipoleIntegrals> as_ints) {
+        as_mp_ints_ = as_ints;
+    }
+
```
fevangelista, line 145:
```suggestion
    void set_active_multipole_integrals(std::shared_ptr<ActiveMultipoleIntegrals> as_mp_ints) {
        as_mp_ints_ = as_mp_ints;
    }

```

forte/helpers/helpers.cc
```diff
@@ -138,6 +139,84 @@ std::pair<double, std::string> to_xb(size_t nele, size_t type_size) {
     return out;
 }
 
+void matrix_transpose_in_place(double* data, const size_t m, const size_t n) {
+    int nthreads = std::min(omp_get_max_threads(), int(m > n ? n : m));
+    std::vector<double> tmp(nthreads * (m > n ? m : n));
+    double* tmp_ptr = tmp.data();
```
fevangelista, line 145: I am not a big fan of the raw pointer here. Can this be avoided?

lcyyork: This function rearranges the data in memory, which is somewhat dangerous. I use it to trick ambit to do in-place transpose of 3-index integrals. Forte uses Qpq order but a more convenient order would be pqQ for DF-PT2. Anyways, I think the raw pointer is more general, but I guess vector<double> works fine as well.

forte/helpers/helpers.h
```diff
@@ -182,10 +185,34 @@ void apply_permutation_in_place(std::vector<T>& vec, const std::vector<std::size
...
+void matrix_transpose_in_place(double* data, const size_t m, const size_t n);
+
+void push_to_psi4_env_globals(double value, const std::string& label);
+
 namespace math {
 /// Return the number of combinations of n identical objects
 size_t combinations(size_t n, size_t k);
 
+/// Compute the greatest common divisor
+inline int gcd(int a, int b) {
```
fevangelista, line 206: Any good reason for not using the std::gcd functions (https://en.cppreference.com/w/cpp/numeric/gcd)?

forte/integrals/cholesky_integrals.cc
```diff
@@ -143,12 +143,22 @@ double** CholeskyIntegrals::three_integral_pointer() { return ThreeIntegral_->po
...
 ambit::Tensor CholeskyIntegrals::three_integral_block(const std::vector<size_t>& A,
                                                       const std::vector<size_t>& p,
-                                                      const std::vector<size_t>& q) {
-    ambit::Tensor ReturnTensor =
-        ambit::Tensor::build(tensor_type_, "Return", {A.size(), p.size(), q.size()});
-    ReturnTensor.iterate([&](const std::vector<size_t>& i, double& value) {
-        value = three_integral(A[i[0]], p[i[1]], q[i[2]]);
-    });
+                                                      const std::vector<size_t>& q,
+                                                      const std::string& order) {
```
fevangelista, line 147: If this function is likely to be called only on the C++ side, why not make order an enum?

fevangelista:
This is a nice PR! I left a few comments but I'd like to go over the PR again once you had a chance to add tests. Ny changing the formulae will be loose support of the original implementation reported in the paper? How big are the differences that you see in the test cases? This would be useful information to put in the PR description.

forte/helpers/helpers.cc
lcyyork: If you are referring line 145, I think it is the most convenient way for lines like 155 and 160.

forte/helpers/helpers.h
lcyyork: Extended GCD is needed for the more optimized in-place matrix transpose algorithm. But currently, the numeric one should be good.

forte/integrals/cholesky_integrals.cc
lcyyork: Yes, it is more likely to be called on the C++ side. Let me make it an enum.

forte/helpers/helpers.cc
lcyyork: Changed to iterator.

forte/integrals/cholesky_integrals.cc
lcyyork: Changed.

lcyyork:
Note sure why PCI-3 failed, but it passes on my computer.

forte/integrals/integrals_psi4_interface.cc
```diff
@@ -369,7 +369,28 @@ void Psi4Integrals::rotate_mos() {
...
+
+    auto Ca_ao = std::make_shared<psi::Matrix>("Ca_AO", nao, nmo_);
+
+    // Transform from the SO to the AO basis
+    for (int h = 0, index = 0; h < nirrep_; ++h) {
+        auto nso = nsopi_[h];
+        if (nso == 0)
+            continue;
+
+        for (int i = 0, nmo_this = nmopi_[h]; i < nmo_this; ++i) {
```
huangm29, line 384: What is the reason we can remove 'resort' from here? Is this the correct way to do the transformation between two sets of Ca?

lcyyork: Now I use Pitzer order here. The `resort` was to make the MO dipole integrals in QT order. This was needed because `Ca_subset("AO")` will return QT order. If the goal is simply remove the symmetry, this is the way to go. It might be tempting to use `remove_symmetry` but it actually transforms both rows and columns and not suitable for removing symmetry of `Ca`.

These several lines of the code are repetitively used in DF integrals and I just collect them here for clarity.

huangm29 (approved):

forte/helpers/helpers.cc
```diff
@@ -138,6 +139,84 @@ std::pair<double, std::string> to_xb(size_t nele, size_t type_size) {
...
 }
 
+void matrix_transpose_in_place(std::vector<double>& data, const size_t m, const size_t n) {
+    int nthreads = std::min(omp_get_max_threads(), int(m > n ? n : m));
+    std::vector<double> tmp(nthreads * (m > n ? m : n));
+    auto tmp_begin = tmp.begin();
+
+    int c = std::gcd(m, n);
+    int a = m / c;
+    int b = n / c;
```
lcyyork, line 149:
```suggestion
    auto c = std::gcd(m, n);
    auto a = m / c;
    auto b = n / c;
```

forte/helpers/helpers.cc
lcyyork: I thought to put it in Psi4 because in-place matrix transposition is not available. I think there are better algorithm out there (e.g., https://onlinelibrary.wiley.com/doi/10.1002/cpe.5071) but the current one does the job.

forte/integrals/integrals.cc
```diff
@@ -181,6 +181,8 @@ const psi::Dimension& ForteIntegrals::ncmopi() const { return ncmopi_; }
 
 size_t ForteIntegrals::ncmo() const { return ncmo_; }
 
+std::vector<size_t> ForteIntegrals::cmotomo() const { return cmotomo_; }
```
fevangelista, line 184:
```suggestion
const std::vector<size_t>& ForteIntegrals::cmotomo() const { return cmotomo_; }
```

fevangelista: I like the fact that by doing it this way we either just pass a pointer or give the options to copy the data.

forte/integrals/integrals.h
```diff
@@ -66,6 +66,14 @@ enum class IntegralSpinRestriction { Restricted, Unrestricted };
  */
 enum IntegralType { Conventional, DF, Cholesky, DiskDF, DistDF, Custom };
 
+/**
+ * @brief The order of three-index integrals when calling
+ *
+ * In Forte, the auxiliary index is the first index (i.e., Qpq). However,
```
fevangelista, line 72:
```suggestion
 * In Forte, the auxiliary index is by default the first index (i.e., Qpq). However,
```

forte/mrdsrg-spin-integrated/master_mrdsrg.cc
```diff
@@ -450,8 +450,8 @@ void MASTER_DSRG::init_dm_ints() {
         dm_[i] = BTF_->build(tensor_type_, "Dipole " + dm_dirs_[i], spin_cases({"gg"}));
     }
 
-    std::vector<psi::SharedMatrix> dm_a = ints_->mo_dipole_ints(true, true);
-    std::vector<psi::SharedMatrix> dm_b = ints_->mo_dipole_ints(false, true);
+    std::vector<psi::SharedMatrix> dm_a = ints_->mo_dipole_ints();
+    std::vector<psi::SharedMatrix> dm_b = dm_a;
```
fevangelista, line 454: Can you please a comment about the logic here?

fevangelista (commented):
Just a few small items to fix. Looks good and looking forward to testing it.

forte/mrdsrg-spin-integrated/master_mrdsrg.cc
lcyyork: Good catch. Because the MO one-body integrals are transformed using only `Ca` so that there is only one set of integrals.

PR #317, opened 2023-03-12
CI restart for MCSCF and MR-DSRG

lcyyork:
## Description
This PR tries to prevent a brand new CI for each MCSCF (or MR-DSRG relax) iteration, but a fixed number of CI iterations are performed.
After the first iteration of MCSCF, `ActiveSpaceSolver` is set to `restart` mode so that the CI initial guesses are from the last iteration.
Every `FCISolver` and `SparseCISolver` now holds a `DavidsonLiuSolver` as an class variable, which may not be memory friendly but it avoids I/O.

TODO: I did some testing of the dynamic sigma build and RDMs algorithms. For the dynamic sigma build, it seems the results are inconsistent with sparse algorithm when using DSRG dressed active integrals. I am uncertain if dynamic sigma build assumes 8-fold symmetry of the two-electron integrals, which DSRG integrals do not have. The dynamic RDM build does not support transition RDMs because the `J` loops start from `I + 1`.

## User Notes
- [x] CI methods stores a DL solver so that it can be restarted.
- [x] `CI_RDMS` accepts `DeterminantSubstitutionLists`.
- [x] `CI_Reference` only accepts `DeterminantHashVec` instead of `std::vector<Determinant>`.

## Checklist
- [x] Documented source code
- [x] Ready to go!


lcyyork:
@fevangelista Hi Francesco, this PR is not yet ready but I think you might want to check a bit before I push too far.

fevangelista:
@lcyyork : let me think a bit about this one. I think that it might make sense to keep the DL solver object inside a FCI code but I was wondering if this should be standardized at the level of the base class.
CI is in essence the combination of:
1) code to enumerate strings/determinants
2) code to compute the sigma vector
3) code to iteratively compute the eigenvalues/eigenvectors using sigma builds
4) code to evaluate properties from the sigma builds
The solver in practice takes care of 1, 2, and 4, and it looks like your PR would package 3 into the solver as well. Correct?

lcyyork:
@fevangelista I guess it is yes to your question.

The issue is that the DL iteration is slow in some cases starting from diagonal H initial guesses. As a result, the tick-tock procedure of MCSCF will be dominant by CI in each macro iteration. A better approach is to 1) use a fixed number of DL iterations for CI in each MCSCF macro iteration (or the residual is lowered by an order of magnitude), 2) save current DL progress, 3) update active-space integrals, and finally 4) resume DL. So there are two ways to achieve this: 1. Keep DL part of CI solver and 2. save DL progress to disk and load up later.

fevangelista:
Absolutely right about dynamic sigma build. I need to investigate.

PR #321, opened 2023-04-25
Add configurations and remove bool*

fevangelista:
## Description
This PR builds some infrastructure for implementing spin adapted CI codes. It also improves the FCI code handling of string substitutions lists.

## User Notes
- [x] Add the `Configuration` class (stores a list of doubly occupied and singly occupied orbitals)
- [x] Removed uses of `bool*` and `vector<bool>`
- [x] Uses the String object (half a determinant) whenever possible
- [x] Improves the speed of generating string substitution lists (🏎️)

## Checklist
- [x] Documented source code
- [x] Ready to go!


fevangelista:
When passes should be merged.

forte/sparse_ci/determinant.h
```diff
@@ -41,6 +42,7 @@ size_t constexpr Norb2 = 2 * Norb;
 
 using String = BitArray<Norb>;
 using Determinant = DeterminantImpl<Norb2>;
+using Configuration = ConfigurationImpl<Norb2>;
```
JonathonMisiewicz, line 45: It doesn't look like `Configuration` is used in this file. Is it included here so that pybind can access it with the existing `#include "sparse_ci/determinant.h"`? If so, mark this resolved.

tests/methods/avas-6/input.dat
```diff
@@ -1,12 +1,12 @@
 #! Test AO Subspace regular expression on 3d orbitals
 
-refscf = -1646.62615747649124 
+refscf = -1646.6261575497
```
JonathonMisiewicz, line 3: Why are these reference values changing?

forte/orbital-helpers/ao_helper.cc
```diff
@@ -105,7 +106,7 @@ void AtomicOrbitalHelper::Compute_AO_Screen(std::shared_ptr<psi::BasisSet>& prim
 void AtomicOrbitalHelper::Estimate_TransAO_Screen(std::shared_ptr<psi::BasisSet>& primary,
                                                   std::shared_ptr<psi::BasisSet>& auxiliary) {
     Compute_Psuedo_Density();
-    MemDFJK jk(primary, auxiliary);
+    MemDFJK jk(primary, auxiliary, psi::Process::environment.options);
```
JonathonMisiewicz, line 109: Why is this needed?

JonathonMisiewicz (commented):
Just to be absolutely clear about this: this PR _adds_ the Configuration class. Use of it in Forte is _planned_ but outside the scope of this PR.

Do I have that right?

fevangelista:
Correct. Here I am just adding the class without using it.

tests/methods/avas-6/input.dat
fevangelista: Good question. The first one is pure SCF from Psi4.

forte/orbital-helpers/ao_helper.cc
fevangelista: This is due to a recent Psi4 PR (#2848) where an option object is required in the initializer. I think the psi4 solution is not ideal and breaks "the C++ side of psi4 should be a library" principle. So this is just a patch. It turns out we don't seem to use AtomicOrbitalHelper anyway.

forte/sparse_ci/determinant.hpp
```diff
@@ -108,6 +108,17 @@ template <size_t N> class DeterminantImpl : public BitArray<N> {
             set_beta_bit(p, occupation_b[p]);
     }
 
+    /// Construct the determinant from an occupation vector that
+    /// specifies the alpha and beta strings.  occupation = [Ia,Ib]
+    explicit DeterminantImpl(const BitArray<nbits_half>& Ia, const BitArray<nbits_half>& Ib) {
+        for (size_t n = 0; n < nwords_half; n++) {
```
JonathonMisiewicz, line 114: Can this constructor not call `set_str`? The logic here is just a call to that function.

JonathonMisiewicz (commented):
Next "big-picture" question: What is the difference between a Configuration and a Determinant?

forte/sparse_ci/determinant.hpp
fevangelista: Good point!

tests/pytest/determinant/test_dets.py
```diff
@@ -190,3 +190,29 @@ def test_det_exciting():
     d7 = det("000")
     assert d7.gen_excitation([], [1, 2], [], [1, 0]) == -1.0
     assert d7 == det("-2+")
+
+
+def test_det_symmetry():
+    """Test class constructors"""
+    if forte.Determinant.norb() >= 128:
```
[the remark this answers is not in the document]
fevangelista: If forte is compiled to support determinants with only 64 orbitals (the default)

PR #345, opened 2023-09-21
TD-DSRG

shuhangli98:
## Overview
This PR is for TD-DSRG, which adds dynamical correlation to TD-ACI by using ACI to propagate the DSRG similarity-transformated Hamiltonian in the active space. 

## Algorithm
HF -> ACI -> DSRG -> Referece relax -> TD-ACI (with relaxed CASCI wavefunction and effective Hamiltonian)

## Task list
- [x] Implement the driver on the Python side
- [x] Change the TDCI class on the C++ side to take in arbitrary PQ spaces/vectors instead of having to first run one ACI calculation.
- [ ] Brian to validate test cases with standalone spinorbital Python code.
- [ ] Enable state-averaging


brianz98:
Looks good! I will generate some TD-CASCI data from my code tomorrow and validate the test case.

PR #352, opened 2023-10-14
Restart CI in MCSCF

fevangelista:
## Description
Add automatic restart of FCI and DETCI computations from the same `ActiveSpaceSolver` object. A few findings:
- `set_read_initial_guess` and `set_die_if_not_converged` don't seem to be necessary in this implementation. @lcyyork should provide feedback.
- setting the maximum number of iterations to 15 seems unnecessary, since for most test cases only the first step takes a lot of iterations.
- It would be nice to have one test case that specifically tests the restart
- Selected CI codes are not restarted. Two things complicate their restart: 1) the code might not be designed for restart, and 2) the possibility that determinants change in number or just ordering complicates the restart procedure (some form of mapping function might be necessary). Future PRs might handle this in the future.
- This code implements some of the changes in PR #317. What features are missing that still need to be implemented?

## User Notes
- [x] Implemented via the new Davidson-Liu solver
- [x] Improved handling of small determinantal spaces
- [x] Add normalization of correction vectors (as implemented in the older code), which is important to be able to converge the residual to very small values (e.g. less than 1.0e-8).  

## Checklist
- [x] Documented source code
- [x] Ready to go!


forte/casscf/mcscf_2step.cc
```diff
@@ -441,23 +460,17 @@ bool MCSCF_2STEP::is_single_reference() {
...
-                                     const std::tuple<int, double, double, bool, bool>& params) {
-    auto state_nroots_map = to_state_nroots_map(state_weights_map_);
-    as_solver = make_active_space_solver(ci_type_, state_nroots_map, scf_info_, mo_space_info_,
-                                         fci_ints, options_);
-
-    int print;
-    double e_conv, r_conv;
-    bool read_wfn_guess, dump_wfn;
-    std::tie(print, e_conv, r_conv, read_wfn_guess, dump_wfn) = params;
+double MCSCF_2STEP::diagonalize_hamiltonian(std::shared_ptr<ActiveSpaceSolver>& as_solver,
```
lcyyork, line 463: There is a bug at line 419 `auto nclosed_electrons = 2 * mo_space_info_->size("INACTIVE_DOCC");` where the 2 should be removed.

huangm29 (approved):

PR #358, opened 2023-11-17
Debugging cluster

huangm29:
## Description
A PR to fix some problems compiling forte using GCC compiler 13 on the cluster

## User Notes
- [ ] Add header files such as `#include<cmath>` or `#include <memory>` for compilation
- [ ] Changing `zheev` to `zheev_` in `sci/tdci.cc` so it works with GCC complier
- [ ] Fix a bug in `tests/pytest/determinant/test_string.py`

## Checklist
- [ ] Ready to go!


marink2 (commented):
This is already fixed in pull request 357

PR #372, opened 2023-12-14
fix casscf-1 c1

lcyyork:
## Description
Add screening to remove small gradients (< 1.0e-12) for test case casscf-1.

## User Notes
- [ ] Features added

## Checklist
- [ ] Documented source code
- [x] Ready to go!


fevangelista:
I think this makes sense as a solution, but I would add that in general it is good to have a MCSCF that can break symmetry (if this is what is happening). I did not have time to look at this yet, but I would like to see what happens to the final orbitals when this patch is not applied. Let me put this on hold for a second.

PR #377, opened 2024-01-29
FCI and GENCI initial core guess determinants for Davidson-Liu solver

marink2:
## Description
PR incorporates the feature to run a FCI (or GENCI) calculation(s) with core-excited determinants as the initial guess space for the Davidson-Liu solver. Core-excited determinants include determinant strings with single- and none-occupation on the first position.

## User Notes
- [x] Features added
- [ ] Changes to compilation (if any)

## Checklist
- [x] Added/updated tests of new features and included a reference `output.ref` file
- [x] Removed comments in code and input files
- [x] Documented source code
- [x] Checked for redundant headers
- [x] Checked for consistency in the formatting of the output file
- [x] Documented new features in the manual
- [x] Ready to go!

PR #379, opened 2024-02-02
Spectra file dump feature

marink2:
## Description

## User Notes
- [ ] Features added

## Checklist
- [ ] Documented source code
- [ ] Ready to go!

### DIFF
--- a/docs/source/options.rst
+++ b/docs/source/options.rst
@@ -739,7 +739,7 @@ Default value: 0.25
 
 **TRANSITION_DIPOLES**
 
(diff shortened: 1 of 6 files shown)


huangm29 (changes requested):
- Please update the User Notes as #377 
- Please add some test functions
- Do you think that the spectra should be dumped when "transition_dipoles" is used? 
- What if one says "transition_dipoles" but only have only one state? I know you should not do that, but it is a possible edge case.

PR #384, opened 2024-04-17
Reset DIIS in DSRG

lcyyork:
## Description
This PR made tweaks to improve the convergence of LDSRG(2).

## User Notes

1. The DIIS subspace is reset when the convergence of cluster amplitudes gets stuck. The criterion for "stuck" is based on the ratio of subsequent amplitude RMS when the RMS is sufficiently small (< 4.0e-3).
2. For the iteratively relaxed version, the convergence of the DSRG amplitudes is no longer tested along the reference relaxation procedure. This is in the same spirit of the two-step MCSCF as there is no need to converge either DSRG or CI in the middle. This change would not affect anything if each DSRG Hamiltonian build is converged within the given number of iterations (controlled by `DSRG_MAXITER`). Essentially, the user should set `DSRG_MAXITER` to smaller number (e.g., 10-20 for systems like Cr2) to achieve speedups for reference relaxation.

## Checklist
- [x] Ready to go!


forte/proc/dsrg.py
```diff
@@ -183,6 +183,7 @@ def dsrg_setup(self):
 
         if self.solver_type in ["SA-MRDSRG", "SA_MRDSRG"]:
             self.dsrg_solver.set_Uactv(self.Ua)
+            self.dsrg_solver.set_die_if_not_converged(self.relax_ref != "ITERATE")
```
fevangelista, line 186: Can you add a comment above to explain the logic of this test?

forte/api/forte_python_module.cc
```diff
@@ -391,6 +391,8 @@ PYBIND11_MODULE(_forte, m) {
              "Set the map from state to the weights of all computed roots")
         .def("set_read_cwd_amps", &SADSRG::set_read_amps_cwd,
              "Set if reading amplitudes in the current directory or not")
+        .def("set_die_if_not_converged", &SADSRG::set_die_if_not_converged,
```
fevangelista, line 394: I see what you are doing. I think that this kind of logic should live on the python side to make it more explicit. But for the moment I am OK with it living here.

forte/mrdsrg-spin-adapted/sa_ldsrg2.cc
```diff
@@ -176,7 +190,7 @@ double SA_MRDSRG::compute_energy_ldsrg2() {
     dump_amps_to_disk();
 
     // fail to converge
-    if (!converged) {
+    if ((!converged) and die_if_not_converged_) {
         clean_checkpoints(); // clean amplitudes in scratch directory
         throw psi::PSIEXCEPTION("The MR-LDSRG(2) computation does not converge.");
```
fevangelista, line 195: Can you use a standard runtime exception instead of the psi specific one?

fevangelista (changes requested):
Thanks for this improvement. I left some minor suggestions.

forte/api/forte_python_module.cc
lcyyork: I agree. I moved the convergence test to python side.

fevangelista:
It looks like there is a small bug. Please push a fix.

fevangelista (approved):

PR #385, opened 2024-04-19
Block2 DMRG Solver

hczhai:
## Description
1. Added the ``block2`` DMRG solver.
2. Fixed a bug in ``RDMsSpinDependent::SF_G2()``.
3. Added ``from __future__ import annotations`` in ``forte/data.py`` to make it work with Python 3.7.

## User Notes
- [x] Features added

- Added a DMRG solver interfaced with ``block2``. Minimal required ``block2`` version is 0.5.3rc7. Installation and usage documentation can be found in: https://block2.readthedocs.io/en/latest/user/forte.html.

- [x] Changes to compilation

- Added a compilation option ``ENABLE_block2``. Default is ``OFF``.

## Checklist
- [x] Documented source code
- [x] Ready to go!


lcyyork (commented):
This is a great feature to have. May I ask for a test case and reference output?

fevangelista:
Thank you for adding this new feature! I'd like to give it some testing before committing it. At some point, the interface will have to change to use the new module system, but that can wait until the interface design is settled.

lcyyork:
In two-step MCSCF, there is a mismatch between the DMRG energy and that reconstructed using DMRG RDMs unless they are in the same canonicalization. Line 260 of `mcscf_2step.cc` may be changed to:
```
bool skip_de_conv = (ci_type_.find("DMRG") != std::string::npos or
                             ci_type_.find("BLOCK2") != std::string::npos);
```

hczhai:
> Line 260 of mcscf_2step.cc may be changed to:

Thanks for pointing this out. This is now added.

fevangelista:
I'll merge this after the last PR changes pass. Let me know if there are other reasons to hold this off.

lcyyork (approved):

forte/dmrg/block2_dmrg_solver.cc
```diff
@@ -0,0 +1,735 @@
...
+            bond_dims.push_back(sweep_bond_dims[i]);
+            noises.push_back(sweep_noises[i]);
+            davidson_tols.push_back(sweep_davidson_tols[i]);
+        }
+    }
+
+    if (n_total_sweeps == 0)
+        n_total_sweeps = n_total_cnt;
+
+    bool read_initial_guess = options_->get_bool("READ_ACTIVE_WFN_GUESS");
```
lcyyork, line 370: I wonder is there a `dump_wfn` implementation available? It seems to me that every DMRG part of the two-step MCSCF procedure always starts from a random initial guess, which leads to oscillating energies even when the orbitals are well converged.

hczhai: The MPS is automatically stored in scratch storage so there is no need to explicitly dump the MPS. One only needs to set this ``READ_ACTIVE_WFN_GUESS`` option (when it is close to convergence) to reuse the MPS from a previous MCSCF step.

fevangelista:
It looks like this PR needs to be changed to address the new modifications in Forte. @hczhai could you take a look? I would be happy to help in case some of the changes are confusing.

fevangelista:
Thank you so much for this contribution @hczhai!

hczhai:
Thanks for merging this! @lcyyork helped a lot in testing and improving this interface.

Actually I have a small additional change https://github.com/hczhai/forte/commit/df179ecb5fd90839f9dcfcb9358c895b15f82fe2 to prevent some future accidental temporary change in block2 source code from breaking the testing pipelines here.

PR #401, opened 2024-08-08
Fix failing tests due to recent Psi4 update

brianz98:
## Description
A recent Psi4 PR (https://github.com/psi4/psi4/pull/3138) fixed the SAD initial guess algorithm, and as a result, some RHF-based tests are landing on (very) slightly different energies, which should cause us no concern. Most CASSCF-based tests are not affected.  This PR fixes the failling tests by updating the reference energies.


tests/methods/aci-18/input.dat
```diff
@@ -2,7 +2,7 @@ import forte
 
 e_scf = -39.72595399459846
 en_ref0 = -39.805891312415
-en_ref1 = -28.917284957659
+en_ref1 = -28.91728496215
```
fevangelista, line 6: Please replace `#  basis cc-pcvdz` with `e_convergence 10` to ensure we converge the SCF procedure tightly.

tests/methods/df-dsrg-mrpt2-1/input.dat
```diff
@@ -2,8 +2,7 @@
 #! Generated using commit GITCOMMIT
 import forte
 
-refmcscf    = -15.569761360884
-refdsrgpt2 =  -15.613389000040760
+refdsrgpt2 =  -15.61338900002254
```
fevangelista, line 5: Please keep the old value but change line 72 to
```
compare_values(refdsrgpt2,variable("CURRENT ENERGY"),9,"DSRG-MRPT2 energy")
```

tests/methods/mp2-nos-1/input.dat
```diff
@@ -1,5 +1,5 @@
 # Test RMP2 natural orbitals
-mp2_corr_ref = -1.802862266979274
+mp2_corr_ref = -1.80286362
```
fevangelista, line 2: If you add `e_convergence 12` after `d_convergence 6` the energy will still match, but there will be a mismatch with the 1RDM. So the problem here is that the test is not well defined. So we need to ensure that SCF is converged tightly first by setting `e_convergence 12`.

fevangelista (changes requested):
Left some minor corrections for you

PR #402, opened 2024-08-28
Enable adaptive RSC threshold

brianz98:
## Description
Use a looser recursive single commutator (RSC) threshold for MR-LRDSRG(2) when delta E is large, and tighter as the energy converges.

## User Notes
- [x] Enable adaptive threshold for spin-adpated hbar build and sequential hbar build

## Checklist
- [ ] Documented source code
- [ ] Ready to go!

### DIFF
--- a/forte/mrdsrg-spin-adapted/sa_ldsrg2.cc
+++ b/forte/mrdsrg-spin-adapted/sa_ldsrg2.cc
@@ -87,24 +87,26 @@ double SA_MRDSRG::compute_energy_ldsrg2() {
 
     // start iteration
     for (int cycle = 1; cycle <= maxiter_; ++cycle) {
+        double Edelta = 0.0;
         // use DT2_ as an intermediate used for compute Hbar
         DT2_["ijab"] = 2.0 * T2_["ijab"];
         DT2_["ijab"] -= T2_["ijba"];
 
         // compute Hbar
+        double rsc_conv = rsc_conv_adapt_ ? get_adaptive_rsc_conv(cycle, Edelta) : rsc_conv_;
         local_timer t_hbar;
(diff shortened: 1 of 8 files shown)


docs/source/methods/dsrg.rst
```diff
@@ -489,6 +489,27 @@ The threshold of considering the BCH expansion converged based on the recursive
 * Type: double
 * Default: 1.0e-12
 
+**DSRG_ADAPTIVE_RSC**
+
+Whether to use an adaptive threshold for the recursive single commutator approximation.
+
+* Type: boolean
+* Default: false
```
lcyyork, line 497: Does the adaptive rsc change the LDSRG(2) results? If not, and it speed up things, I prefer to make it `true` by default.

lcyyork (commented):
Thanks for the PR! I wonder if this adaptive cutoff changes the final answer and how much speedup can we gain. Would you mind to put some head-to-head comparisons in the Description section?

brianz98:
> Thanks for the PR! I wonder if this adaptive cutoff changes the final answer and how much speedup can we gain. Would you mind to put some head-to-head comparisons in the Description section?

I've just added a benchmark calculation. Instead of calling the Hbar build routine 144 times, the adaptive version calls it 81 times (using the default settings), and the final energies agree to 1e-8 Eh.

lcyyork (approved):
This seems a great improvement! Thanks for making the comparison.

docs/source/methods/dsrg.rst
brianz98: I think so too. I'll run all of the ldsrg2 tests and see if any of them fail.

PR #410, opened 2024-09-10
Fix #409 : Managing Psi Dependencies and Options

### LINKED ISSUE #409: `psi::Process::environment.molecule()` Incorrect
A few places in the code, Forte tries to access the current molecule via `psi::Process::environment.molecule()`. This may not refer to the molecule of interest.

In some cases, Psi4 never actually sets the molecule, causing a segfault in `state_info.cc::make_state_info_from_psi`. Example:
```
import forte

psi4.schema_wrapper.run_qcschema({"id": None, "schema_name": "qcschema_input", "schema_version": 1, "molecule": {"schema_name": "qcschema_molecule", "schema_version": 2, "validated": True, "symbols": ["C", "C", "C", "C", "C", "C", "H", "H", "H", "H"], "geometry": [0.0, -2.539223068036, 0.0, 0.0, 2.539223068036, 0.0, -2.281614274862, -1.347763537454, 0.0, 2.281614274862, -1.347763537454, 0.0, 2.281614274862, 1.347763537454, 0.0, -2.281614274862, 1.347763537454, 0.0, -4.07273674735, -2.318840739012, 0.0, 4.07273674735, -2.318840739012, 0.0, 4.07273674735, 2.318840739012, 0.0, -4.07273674735, 2.318840739012, 0.0], "name": "pbenzyne", "molecular_charge": 0.0, "molecular_multiplicity": 1, "masses": [12.0, 12.0, 12.0, 12.0, 12.0, 12.0, 1.00782503223, 1.00782503223, 1.00782503223, 1.00782503223], "real": [True, True, True, True, True, True, True, True, True, True], "atom_labels": ["", "", "", "", "", "", "", "", "", ""], "atomic_numbers": [6, 6, 6, 6, 6, 6, 1, 1, 1, 1], "mass_numbers": [12, 12, 12, 12, 12, 12, 1, 1, 1, 1], "fragments": [[0, 1, 2, 3, 4, 5, 6, 7, 8, 9]], "fragment_charges": [0.0], "fragment_multiplicities": [1], "fix_com": True, "fix_orientation": True, "fix_symmetry": "c1", "provenance": {"creator": "QCElemental", "version": "0.28.0", "routine": "qcelemental.molparse.from_string"}, "extras": {}}, "driver": "energy", "model": {"method": "forte", "basis": "cc-pcvdz"}, "keywords": {"ACTIVE": [8], "DOCC": [20], "D_CONVERGENCE": 1e-10, "E_CONVERGENCE": 1e-10, "G_CONVERGENCE": "GAU_VERYTIGHT", "MAXITER": 300, "MCSCF_E_CONVERGENCE": 1e-10, "MCSCF_MAXITER": 500, "MCSCF_R_CONVERGENCE": 1e-10, "MCSCF_TYPE": "DF", "OPT_COORDINATES": "BOTH", "REFERENCE": "RHF", "RESTRICTED_DOCC": [16], "R_CONVERGENCE": 1e-10, "SCF_TYPE": "DF", "forte__scf_type": "DF", "forte__e_convergence": 1e-10, "forte__d_convergence": 1e-10, "forte__active_space_solver": "FCI", "forte__multiplicity": 1, "forte__restricted_docc": [16], "forte__active": [8], "forte__int_type": "DF", "forte__r_convergence": 1e-10, "forte__mcscf_maxiter": 500, "forte__mcscf_e_convergence": 1e-10, "forte__reference": "RHF", "forte__maxiter": 200, "forte__basis": "CC-PCVDZ", "forte__df_basis_mp2": "CC-PVDZ-JKFIT", "function_kwargs": {}, "PARENT_SYMMETRY": "C1"}, "protocols": {"stdout": True}, "extras": {"psiapi": True, "wfn_qcvars_only": True}, "provenance": {"creator": "QCElemental", "version": "0.28.0", "routine": "qcelemental.models.results"}})
```

In other cases, this will refer to the wrong molecule. For example, Forte does SCF on the _neutral_ and ACI on the _cation_ of the following input.

```
import forte

molecule waterNeutral {
O
H 1 1.0
H 1 1.0 2 104.5
}

molecule waterCation {
1 2
O
H 1 1.0
H 1 1.0 2 104.5
}

set basis sto-6g

set forte {
   active_space_solver aci
}

energy('forte', molecule=waterNeutral)
```

JonathonMisiewicz:
## Description
This PR aims to fix issue #409 : dependence on `psi::Process::environment.molecule()` cam make Forte behave badly. This PR is a bit delicate, so @fevangelista approval is **mandatory** before I'm comfortable merging this in.

There are three general classes of fixes necessary.

1. `make_state_info_from_psi` has read the molecule from Psi's globals from the beginning, and it needs to get that information from somewhere else. Based on the use case of this function, it seemed best to rename it to `make_state_info_from_options` and have it get everything from the ForteOptions object and the point group. See https://github.com/evangelistalab/forte/blob/9b468bd7f3250ed3bc9545c07019c9942a95746e/forte/base_classes/active_space_solver.cc#L422 for a function that really shouldn't have to assume the existence of a Psi molecule. `prepare_forte_objects` is now responsible for moving wavefunction information to the object. To avoid contaminating globals, the OptionsFactory has been taught to construct data that is a copy of the global options rather than a reference... This may not be the best way to solve this particular problem, and I'm open to counter-suggestions. The best option I could think of was to have separate `make_state_info_from_options` and then a `make_state_info_from_psi` that explicitly needs a Psi molecule. Ensuring the correct one was called seemed like a headache, but I didn't look much into it.
2. Several functions used the molecule name as part of a file or directory name. This has been replaced with the active space string. For active space methods, I've also added the state string. DynamicCorrelationSolver objects can't assume one particular state, based on my reading of the code, so they can't do the same.
3. Appeals to the environment molecule for nuclear dipole have been replaced with calls to `MultipoleIntegrals`, which uses the molecule on the wavefunction. This is still, unfortunately, dependent on Psi, but I could see no way to remove it without a lot of trouble.

## User Notes
- [x] Some temporary file names have changed.
- [x] State information must now be mentioned in options. Forte should automatically handle this for you. If you see a problem, file an issue and ping @JonathonMisiewicz. 

## Checklist
- [x] Tests run locally - everything that isn't block2-related passed
- [x] Ready to go!


fevangelista:
Thank you for looking into this issue. I have to think more about this, but I was thinking that perhaps we should not rely on `Process` anymore to get the molecule object; it should be passed directly and held by Forte somewhere.

JonathonMisiewicz:
I considered Forte storing the `psi4.Molecule`, but the problem I could never answer was what to do for PySCF integrals.

fevangelista:
Thanks for this PR. I left some suggestions for improvements. One aspect that I think @lcyyork  will need to comment on is changes like:
```
-   psi::Process::environment.molecule()->name();
+   std::to_string(mo_space_info_->size("ACTIVE"));
```
What is a good way to generate a molecule/computation label? This seems to be used to denote files. I don't have a good suggestion at the moment.

JonathonMisiewicz:
Since I'm at Virginia Tech now and you're already open to Forte storing a molecule, might I suggest storing a molecule [in a QCSchema format](https://molssi.github.io/QCElemental/model_molecule.html#api)? It naturally has a name field, Psi4 most definitely supports QCSchema, PySCF probably does, Forte already likes communicating in JSON, and we're tying ourselves to one format _minimally_ because it's a community standard.

fevangelista:
I'm open to storing the molecule in the QCSchema. It does have some restrictions (e.g., truncation of positions to 8 digits). However, the biggest concern is that it is all on the Python side, so every time we need to access the molecule from the C++ side, we would have to do some translation. Hopefully, this can be minimized. We would also want to ignore the charge/multiplicity coming from the QCSchema molecule since it has a default of 0 and 1, which, although convenient, in my opinion, opens up the door to user errors.

lcyyork:
> Thanks for this PR. I left some suggestions for improvements. One aspect that I think @lcyyork will need to comment on is changes like:
> 
> ```
> -   psi::Process::environment.molecule()->name();
> +   std::to_string(mo_space_info_->size("ACTIVE"));
> ```
> 
> What is a good way to generate a molecule/computation label? This seems to be used to denote files. I don't have a good suggestion at the moment.

I do prefer the name of molecule indeed.

For the issue, would `activate(mol)` give the desired behavior?

lcyyork:
> Thanks for this PR. I left some suggestions for improvements. One aspect that I think @lcyyork will need to comment on is changes like:
> 
> ```
> -   psi::Process::environment.molecule()->name();
> +   std::to_string(mo_space_info_->size("ACTIVE"));
> ```
> 
> What is a good way to generate a molecule/computation label? This seems to be used to denote files. I don't have a good suggestion at the moment.

I do prefer the name of molecule indeed.

For the issue, would `activate(mol)` give the desired behavior?

JonathonMisiewicz:
> For the issue, would `activate(mol)` give the desired behavior?

No. I posted _two_ problem input files. That would fix Example 2 but not Example 1.

JonathonMisiewicz:
> For the issue, would `activate(mol)` give the desired behavior?

I haven't tested this, but I imagine it would fix some cases but not all. Your proposal assumes there _is_ a Psi molecule. That would fix the examples I posted, but you couldn't use this with PySCF molecules, which Francesco wants to support.

fevangelista:
I think we should go with the general solution rather than the hack. At least this moves us closer to interoperability with other software.

fevangelista (approved):

PR #411, opened 2024-09-13
4-RDM

shuhangli98:
Implements spin-dependent 4-RDM.

### DIFF
--- a/forte/genci/genci_string_lists.cc
+++ b/forte/genci/genci_string_lists.cc
@@ -423,4 +423,16 @@ std::vector<H3StringSubstitution>& GenCIStringLists::get_beta_3h_list(int h_I, s
     return beta_3h_list[I_tuple];
 }
 
+std::vector<H4StringSubstitution>& GenCIStringLists::get_alfa_4h_list(int h_I, size_t add_I,
+                                                                      int h_J) {
+    std::tuple<int, size_t, int> I_tuple(h_I, add_I, h_J);
+    return alfa_4h_list[I_tuple];
+}
+
+std::vector<H4StringSubstitution>& GenCIStringLists::get_beta_4h_list(int h_I, size_t add_I,
+                                                                      int h_J) {
+    std::tuple<int, size_t, int> I_tuple(h_I, add_I, h_J);
+    return beta_4h_list[I_tuple];
+}
+
 } // namespace forte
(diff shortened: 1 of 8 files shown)


forte/genci/genci_vector_rdm.cc
```diff
@@ -125,7 +182,16 @@ std::shared_ptr<RDMs> GenCIVector::compute_rdms(GenCIVector& C_left, GenCIVector
...
         }
+        if (max_rdm_level == 4) {
+            return std::make_shared<RDMsSpinDependent>(g1a, g1b, g2aa, g2ab, g2bb, g3aaa, g3aab,
+                                                       g3abb, g3bbb, g4aaaa, g4aaab, g4aabb, g4abbb,
+                                                       g4bbbb);
+        }
     } else {
+        if (max_rdm_level == 4) {
+            throw std::runtime_error(
+                "Spin-free RDMs of order 4 or higher are not implemented in GenCISolver");
```
fevangelista, line 193: This will take care if the user demands 4RDMs. We should still add the exceptions to the accessors of the SpinFreeRDMs class

forte/genci/string_lists_makers.cc
```diff
@@ -387,6 +387,97 @@ H3List make_3h_list(const StringList& strings, std::shared_ptr<StringAddress> ad
     return list;
 }
 
+H4List make_4h_list(const StringList& strings, std::shared_ptr<StringAddress> addresser,
```
fevangelista, line 390: Can you time this for say, a CAS(10,10) calculation (without computing the 4-RDM) and report it in the PR notes?

fevangelista (changes requested):
Thanks for adding the 4-RDMs. This is a nice addition. We can probably tweak the code for performance but it would just be good to have a reference implementation in Forte. The RDMs tests function takes care of that. What about adding an example with 3a/2b electrons to test that the different spin cases work as expected?

forte/genci/genci_string_lists.cc
```diff
@@ -154,6 +155,8 @@ void GenCIStringLists::startup(std::shared_ptr<MOSpaceInfo> mo_space_info) {
     gas_beta_2h_occupations_ = generate_1h_occupations(gas_beta_1h_occupations_);
     gas_alfa_3h_occupations_ = generate_1h_occupations(gas_alfa_2h_occupations_);
     gas_beta_3h_occupations_ = generate_1h_occupations(gas_beta_2h_occupations_);
+    gas_alfa_4h_occupations_ = generate_1h_occupations(gas_alfa_3h_occupations_);
```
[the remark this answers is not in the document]
brianz98: We'll deal with it in this PR. Commit soon.

PR #414, opened 2024-09-19
Update math_structures.h

lcyyork:
## Description
Somehow line 201 throws an error (`error: invalid operands to binary expression ('std::complex<double>' and 'int')`) on my Linux server using intel compiler.

## Checklist
- [x] Ready to go!


fevangelista:
Thanks for catching this! This is the right way to do it.

PR #418, opened 2024-09-26
Simpleapi

fevangelista:
## Description
This is a big maintenance PR that moves Forte towards a more modular design. The initial goal was to simplify the interface to run GAS computations but that will be left for a future PR.

## User Notes
- [x] Introduces experimental apps and simplified input for defining active spaces (experimental)
- [x] Decouples `MOSpaceInfo` from `ForteOptions` (plus maintenance work in `MOSpaceInfo`)
- [x] Introduces the Observer pattern to link MOs (`SCFInfo`) to the integral class
- [x] Reworks the `Integral` class to use the Observer pattern plus some maintenance work in this class
- [x] Added the `Workflow` module and rewritten the `ObjectsFromPsi4` and `ObjectsUtilPsi4` modules
- [x] Updated tests to use this new interface 

## Checklist
- [x] Documented source code
- [x] Ready to go!


forte/api/orbital_api.cc
```diff
@@ -70,9 +70,9 @@ void export_Localize(py::module& m) {
 void export_SemiCanonical(py::module& m) {
     py::class_<SemiCanonical>(m, "SemiCanonical")
         .def(py::init<std::shared_ptr<MOSpaceInfo>, std::shared_ptr<ForteIntegrals>,
-                      std::shared_ptr<ForteOptions>, bool, bool, bool>(),
-             "mo_space_info"_a, "ints"_a, "options"_a, "inactive_mix"_a, "active_mix"_a,
-             "quiet"_a = false)
+                      std::shared_ptr<ForteOptions>, std::shared_ptr<SCFInfo>, bool, bool, bool>(),
+             "mo_space_info"_a, "ints"_a, "options"_a, "scf_info"_a, "inactive_mix"_a,
```
JonathonMisiewicz, line 74: I'm under the impression that after this PR, the Ints always stores an SCF Info. If so, why is this necessary?

fevangelista: Good question. With this PR, SCFInfo and ForteIntegrals will always be coherent. Meaning that a change in SCFInfo will trigger a transformation of the integrals (or at least flagging the integrals as dirty). This is what the observer pattern helps express. The point of doing this is to make it difficult for programmers to create integrals out of a set of MO coefficients and then have some part of the code modify the integrals or the MO coefficients without the two being synchronized.

PR #419, opened 2024-09-30
Rename EPICTensors -> ndarray. Cleanup code and add tests.

fevangelista:
## Description
This PR is a continuation of the original work on tensor storage and consists mostly of renaming the tensor class and adding better accessors. Note that the commented tests are not yet supported and will drive the development of future PR that continue this work.

User questions:
- What data type should we try to support? Currently available data types are `float`, `double`, `complex<float>`, `complex<double>`. Would support for `int32/64` or `bool` be useful?

## User Notes
- [x] Renamed class to standardize with nanobind (anticipating future adoption)
- [x] Added test cases

## Checklist
- [x] Documented source code
- [x] Ready to go!


forte/helpers/ndarray/ndarray.hpp
```diff
@@ -0,0 +1,399 @@
...
+using namespace pybind11::literals;
+
+constexpr bool ndarray_debug = true;
+
+namespace forte {
+
+/// @brief Class to represent a tensor
+template <typename T> class ndarray {
+    /// @brief Shape of tensor
+    using shape_t = std::vector<size_t>;
```
brianz98, line 25: Mirroring `np.empty` or `np.zeros`, should we change `shape_t` to `std::tuple`?

fevangelista: That's a good suggestion, but unfortunately we don't template over the number of dimensions (should we?), so this has to be a dynamically allocated quantity (std::tuple has compile-time dimension fixed).

brianz98: Good point! I still think front a user's perspective `ndarray([3,4])` looks like it should return a 1-D array of length 2 set to 3 and 4, instead of a 2-D empty array with 12 elements. So it might be worth just doing a pybind type conversion from tuple to list before passing to C++?

fevangelista: I see what you mean. I now have enabled passing the shape via a tuple. The list is disabled. To make the class accept a list with data, I would need to write code to handle arbitrary cases, which is not simple. Since the point of this object is to facilitate Py<->C++ interoperability, it won't reproduce the full extent of capabilities of ndarrays in np. If the user wants to create an array from a list or list of lists of numbers, they can do so in numpy and then create a C++ ndarray from the python array.

PR #427, opened 2024-10-29
Update math_structures.h

lcyyork:
## Description
Intel 2025.0 complains about the `insert` method in `VectorSpaceList`, which is only defined in `VectorSpace`. I think the correct way is to use `add` instead of `insert` for `VectorSpaceList`. Please close this PR if I am wrong. Thanks!

## User Notes
- [ ] Features added

## Checklist
- [ ] Documented source code
- [x] Ready to go!


fevangelista:
The compiler is right. This was never detected because the function is probably not used (and therefore not compiled).

fevangelista (approved):

PR #447, opened 2025-02-11
Fix general CC DIIS

brianz98:
## Description
This PR fixes the DIIS implementation in the general CC / SRCC code. The old implementation saved all iterations instead of the most recent n iterations, and convergence was typically not improved much compared to without DIIS.

## Checklist
- [x] Ready to go!

### DIFF
--- a/forte/modules/general_cc.py
+++ b/forte/modules/general_cc.py
@@ -346,8 +346,8 @@ class DIIS:
             bsol = np.zeros(diis_dim + 1)
             B[-1, -1] = 0.0
             bsol[-1] = -1.0
-            for i in range(len(self.e_diis)):
-                for j in range(i, len(self.e_diis)):
+            for i in range(diis_dim):
+                for j in range(i, diis_dim):
                     B[i, j] = np.dot(np.real(self.e_diis[i]), np.real(self.e_diis[j]))
                     if i != j:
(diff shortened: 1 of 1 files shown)


forte/modules/general_cc.py
```diff
@@ -337,8 +339,8 @@ def update(self, t, t_old):
         self.t_diis.append(t)
         self.e_diis.append(np.subtract(t, t_old))
 
-        diis_dim = len(self.t_diis) - 1
-        if (diis_dim >= self.diis_start) and (diis_dim < len(t)):
+        diis_dim = len(self.t_diis)
```
fevangelista, line 342: Can we use `diis_dim` also in the loop around line 350 as below?
```
            for i in range(diis_dim):
                for j in range(i, diis_dim):
```

fevangelista (approved):

fevangelista:
Thanks for this PR! Can you leave some comment on the observed performance increase (e.g. comparison of number of iterations?). This would be good to have for reference.

brianz98:
> Thanks for this PR! Can you leave some comment on the observed performance increase (e.g. comparison of number of iterations?). This would be good to have for reference.

here is the number of iterations before / after for short tests in the `tests/pytest/srcc` directory
image

PR #448, opened 2025-02-12
Generate truncated hilbert spaces

brianz98:
## Description
This PR is a small modification on the `make_hilbert_space` function that allows one to truncate the total excitation rank relative to the aufbau determinant.

## Checklist
- [x] Documented source code
- [x] Ready to go!

### DIFF
--- a/forte/api/determinant_api.cc
+++ b/forte/api/determinant_api.cc
@@ -164,9 +164,22 @@ void export_Determinant(py::module& m) {
         .def("get_idx", &DeterminantHashVec::get_idx, " Return the index of a determinant");
 
     m.def(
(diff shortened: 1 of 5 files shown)


fevangelista (changes requested):
Can I suggest a slightly different strategy here? Can we have the aufbau determinant passed to the function? That would expand significantly its use case.

brianz98:
> Can I suggest a slightly different strategy here? Can we have the aufbau determinant passed to the function? That would expand significantly its use case.

Thanks for the suggestion! It is now implemented and tested.

brianz98:
@fevangelista having thought about it a little more - do we want to enforce the symmetry of the reference determinant to be the symmetry of generated determinants? For example, what if we want to generate all B1u determinants that can be coupled to the Ag aufbau reference by the dipole operator or some other non totally symmetric operators?

fevangelista (approved):

PR #451, opened 2025-02-19
Add reverse option to `SparseFactExp`

brianz98:
## Description
`SparseFactExp::apply_op/antiherm(oplist, state)` performs the action `... exp(oplist[1]) exp(oplist[0])|state>`, which is expected/conventional behavior for an operator list built iteratively, but is not what one might naively expect otherwise (e.g., if you have a list of singles and then doubles, you might expect the function to apply doubles onto the state first, then singles, etc). This PR adds the optional argument `reverse` to these two functions to do this. It works with the `inverse` argument in the expected way (inverse negates all coefficients, while XOR(inverse,reverse) determines if the list is reversed)

## Checklist
- [x] Documented source code
- [x] Ready to go!


brianz98:
Merged with #450

PR #454, opened 2025-04-04
Make `forte.overlap` respect argument order

brianz98:
## Description
This PR restores the expected behavior of `forte.overlap(a,b) == <a|b>`, where before the ket was whichever one of `a` or `b` that's the shorter vector. This should only impact results where the vectors have complex coefficients. A test has been added to that effect.

## Checklist
- [x] Ready to go!

### DIFF
--- a/forte/helpers/math_structures.h
+++ b/forte/helpers/math_structures.h
@@ -281,7 +281,7 @@ class VectorSpace {
         return result;
     }
 
-    /// @brief Calculate the dot product of two vectors
+    /// @brief Calculate the dot product of two vectors <this|other>
     F dot(const Derived& other) const {
         F result{0};
         bool self_smaller = size() < other.size();
--- a/forte/helpers/math_structures.h
+++ b/forte/helpers/math_structures.h
@@ -284,11 +284,16 @@ class VectorSpace {
     /// @brief Calculate the dot product of two vectors
     F dot(const Derived& other) const {
         F result{0};
+        bool self_smaller = size() < other.size();
         const auto& smaller = size() < other.size() ? elements() : other.elements();
         const auto& larger = size() < other.size() ? other.elements() : elements();
         for (const auto& [e, c] : smaller) {
             if (const auto it = larger.find(e); it != larger.end()) {
-                result += conjugate(c) * it->second;
+                if (self_smaller) {
+                    result += conjugate(c) * it->second;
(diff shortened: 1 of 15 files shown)


forte/helpers/math_structures.h
```diff
@@ -284,11 +284,20 @@ class VectorSpace {
     /// @brief Calculate the dot product of two vectors
```
fevangelista, line 284:
```suggestion
    /// @brief Calculate the dot product of two vectors <this|other>
```

forte/helpers/math_structures.h
```diff
@@ -284,11 +284,20 @@ class VectorSpace {
...
+        if (self_smaller) {
+            for (const auto& [e, c] : smaller) {
+                if (const auto it = larger.find(e); it != larger.end()) {
+                    result += conjugate(c) * it->second;
+                }
+            }
+        } else {
+            for (const auto& [e, c] : smaller) {
+                if (const auto it = larger.find(e); it != larger.end()) {
+                    result += c * conjugate(it->second);
```
fevangelista, line 299:
```suggestion
                    result += conjugate(it->second) * c;
```

fevangelista (approved):

fevangelista:
All looks good to me. There is only a small issue with the cc tests. They seem to be converging to the correct value, but the error has crept beyond 1e-11.

brianz98:
> All looks good to me. There is only a small issue with the cc tests. They seem to be converging to the correct value, but the error has crept beyond 1e-11.

These failures have already been present in main for a couple days now - seems to be introduced either by the aset or the sparse hamiltonian commits. Could it be a change in default tolerances somewhere in the latter?

fevangelista:
> > All looks good to me. There is only a small issue with the cc tests. They seem to be converging to the correct value, but the error has crept beyond 1e-11.
> 
> These failures have already been present in main for a couple days now - seems to be introduced either by the aset or the sparse hamiltonian commits. Could it be a change in default tolerances somewhere in the latter?

I checked the PRs, if you tighten the convergence of the GeneralCC module it matches the original values. The problem is that for GenCC `e_convergence` is set to 1e-10, but the test checks for 1e-11. So the solution is to change the test tolerances to at least 1e-10, but probably 5e-10 would be better. Can you do that in this PR for the examples that fail?

PR #457, opened 2025-06-27
Added function to write integrals that DSRG uses to an FCIDUMP

EmilyHuang2005:
## Description
- Function write_to_fcidump() that writes integrals that DSRG uses to INTDUMP, then stops calculation

## User Notes
- Automatic number of electrons not yet implemented
- Called function under make_dsrg_solver() but need option implemented to determine whether or not it should be called
- FCIDUMP is in PySCF format

### DIFF
--- a/forte/options.yaml
+++ b/forte/options.yaml
@@ -143,6 +143,14 @@ Driver:
     type: bool
     default: False
     help: "Save orbitals to file if true"
+  WRITE_INTDUMP:
+    type: bool
+    default: False
+    help: "Write the ForteIntegral object to INTDUMP file in FCIDUMP format"
+  INTDUMP_NAME:
+    type: str
+    default: "INTDUMP"
+    help: "The name of the INTDUMP file to write to. Used only if WRITE_INTDUMP is true."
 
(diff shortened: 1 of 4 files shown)


brianz98:
I have refactored the function, putting it into `proc/fcidump.py`, and added two forte options: `WRITE_INTDUMP` (boolean) and `INTDUMP_NAME` (str) for calling it before (optionally) entering a DSRG computation. By the way, this makes it possible to only run MCSCF / ASET without having to run a DSRG calculation. Simply remove all DSRG-related keywords like `correlation_solver`, and this _should_ still run. You also need to add a few test cases, e.g. getting INTDUMPs from a 
1. CASCI (`mcscf_reference  false`), 
2. CASSCF (should especially test with `frozen_docc` set - this shouldn't freeze during mcscf, but should give you an INTDUMP file without the frozen orbitals), and 
3. ASET(0)

calculation, and then putting it back into a DSRG calculation by `int_type FCIDUMP`. These are then to be compared to conventional DSRG results without going through the intdump intermediates.

